PR #9, opened 2020-06-04
CVMFS introduction

robberteggermont:


robberteggermont:
How can I remove the first commit from this pull request?

boegel:
@robberteggermont You can, but you don't really need to I think, unless you want to keep the `git` history clean.

It involves running `git rebase --interactive master`, then marking the other commit to be removed, and then repushing with `git push --force`.

Can you clarify why you feel the need to remove the first commit?

The combined proposed changes of both commits look OK to me (see the `Files` tab at https://github.com/EESSI/docs/pull/9/files). It doesn't really matter that you overwrite the changes made in the first with the 2nd commit imho...

robberteggermont:
> Can you clarify why you feel the need to remove the first commit?

Just to keep the history clean, since I dropped everything but the title from the first commit. But I'll leave it like this.

boegel:
Nice and short yet clear, excellent!

Should we also add a link to our own CVMFS repository before we merge this? Something like:

```
Detailed information about how we configure CVMFS is available at https://github.com/EESSI/CVMFS-layer .
```

Once the procedure to set up a Stratum-1/squid proxy and how to mount the filesystem has been fleshed out by @bedroge and tested by others (@trz42 & co), we can also include that more technical info on this page in a subsection.

robberteggermont:
@boegel Sorry, I like my information to the point, don't count on novels from me.;-)

The link has been added.

PR #27, opened 2020-06-17
Don't make the logo so dominant

ocaisa:
... and this way it also scale according to the device

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,4 +1,6 @@
-![EESSI logo](img/logos/EESSI_logo_horizontal.png)
+<p align="center">
+  <img width="60%" src="img/logos/EESSI_logo_horizontal.png" alt="EESSI logo">
+</p>
 
 # Welcome to the EESSI project documentation!
 


docs/index.md
```diff
@@ -1,4 +1,6 @@
-![EESSI logo](img/logos/EESSI_logo_horizontal.png)
+<p align="center">
+  <img width="60%" src="img/logos/EESSI_logo_horizontal.png" alt="EESSI logo">
```
boegel, line 2: Did you test this, does it work in a deployed version of the docs?

boegel: See also https://github.com/mkdocs/mkdocs/issues/991 (not sure if that's fixed)

ocaisa: Nope, didn't test, assumed the CI would that for me.

If there's a related bug, I can just use an absolute link

ocaisa: It rendered fine in the editor

ocaisa: According to [this comment](https://github.com/mkdocs/mkdocs/issues/991#issuecomment-233213340) it should work fine

ocaisa: Tested, works

PR #29, opened 2020-06-17
Drop navigation tabs on top

### LINKED ISSUE #28: Drop "Home" tab
Is there a way to drop the "Home" tab in the header? That introduces a lot of space that is useless unless we add more tabs (and the logo already performs the home function)

ocaisa:
Fixes #28 
If we only have a `Home` tab then that is pretty useless

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -7,11 +7,8 @@ theme:
   name: material
   features:
     - instant
-    - tabs
   logo: img/logos/EESSI.png
   favicon: img/logos/EESSI.png
-  feature:
-    tabs: false
 repo_name: EESSI @ GitHub
 repo_url: https://github.com/EESSI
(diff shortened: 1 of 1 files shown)


ocaisa:
Relevant documentation: https://squidfunk.github.io/mkdocs-material/getting-started/#tabs

mkdocs.yml
```diff
@@ -10,6 +10,8 @@ theme:
     - tabs
   logo: img/logos/EESSI.png
   favicon: img/logos/EESSI.png
+  feature:
+    tabs: false
```
boegel, line 14: just remove the `- tabs` on line 10 instead?

ocaisa: Yup, fixing now

ocaisa:
Tested, works on rendered docs

PR #34, opened 2020-07-14
add link to recording of HPCKP talk

boegel:


ocaisa:
@boegel Can you synch your branch so the CI passes

boegel:
> @boegel Can you synch your branch so the CI passes

done!

PR #43, opened 2020-10-22
Added arm pilot image support.

terjekv:


docs/pilot.md
```diff
@@ -42,7 +42,8 @@ The container image can be used directly by Singularity (no prior download requi
   ```shell
   export EESSI_CONFIG="container:cvmfs2 cvmfs-config.eessi-hpc.org /cvmfs/cvmfs-config.eessi-hpc.org"
   export EESSI_PILOT="container:cvmfs2 pilot.eessi-hpc.org /cvmfs/pilot.eessi-hpc.org"
-  singularity shell --fusemount "$EESSI_CONFIG" --fusemount "$EESSI_PILOT" docker://eessi/client-pilot:centos7-2020.09
+  uname -m |grep -i aarch > /dev/null && export EESSI_DOCKER_ARCH="-arm64v8"
```
boegel, line 45: Note to self: also update client container images so we can just use `$(uname -m)` here without hacking around like this (cfr. https://github.com/EESSI/compatibility-layer/issues/52)

terjekv: Yeah. This was not... pretty. :)

robberteggermont (approved):

PR #55, opened 2021-01-19
add slides + link to recording for EESSI 'behind the scenes' talk

boegel:


bedroge:
What about the questions, where do we put those?

boegel:
Ah, good point. I guess I can copy-paste them in this README...

bedroge (approved):

PR #59, opened 2021-02-02
Add CernVM workshop talk

bedroge:


boegel:
Maybe also include the questions which were raised (where possible with names added)?

This is taken from the shared notes for the CernVM-FS workshop:

```
Q: Are some of the HPC Center in EuroHPC or PRACE
A: Julich and CSCS are examples for large centers (not sure if part of PRACE or EuroHPC)
A: LUMI has shown signs of interest.

Q: Key4HEP would be interested in a GitHub PR based workflow as envisaged by EESSI, interested in collaboration; perhaps the Github action runner can help.
A: Expect the cvmfs ephemeral publish container would help

Q: Rocky Linux already use building based on github PRs
A: Thanks for the comment

Q: Does your tools support multiple versions/compilers of a package in the same build tree?
A: We don’t do that at the moment but modules should in principle support that fine
A: We have multiple versions of OpenFOAM already in EESSI

Q: Interested in continuing the discussion to learn in HSF from EESSI packaging experience
A: Happy to get in touch
```

robberteggermont (approved):

PR #65, opened 2021-03-09
Update list of supported microarchitectures

bedroge:
Also updates the output that you get when sourcing the init script, this was outdated.


docs/pilot.md
```diff
@@ -205,8 +198,14 @@ EESSI currently supports the following HPC applications as well as all their dep
 #### aarch64/arm64
 
   - generic (currently implies `-march=armv8-a` and `-mtune=generic`)
+  - AWS Graviton2
```
ocaisa, line 201: Why AWS? Is is specific to them?

bedroge: Yes, it's being developed by a company that was acquired by Amazon:
https://en.wikipedia.org/wiki/Annapurna_Labs#AWS_Graviton2
AFAIK they're only available in AWS.

ocaisa (approved):

PR #84, opened 2021-12-16
Updated instructions for setting up Stratum 1

rungitta:
This should solve:
https://github.com/EESSI/filesystem-layer/issues/100
https://github.com/EESSI/filesystem-layer/issues/65

### DIFF
--- a/docs/filesystem_layer/stratum1.md
+++ b/docs/filesystem_layer/stratum1.md
@@ -52,7 +52,7 @@ ansible-galaxy role install -r requirements.yml -p ./roles --force
 
 Make sure you have enough space in `/srv` (on the Stratum 1) since the snapshot of the Stratum 0
 will end up there by default. To alter the directory where the snapshot gets copied to you can add
-this variable in `inventory/host_vars/stratum1host`:
+this variable in `inventory/host_vars/<url-or-ip-to-your-stratum1>`:
 
 ```bash
 cvmfs_srv_mount: /srv
--- a/docs/filesystem_layer/stratum1.md
+++ b/docs/filesystem_layer/stratum1.md
@@ -98,7 +98,7 @@ Make sure to include the IP address of your server.
 ## Step 3: Verification of the Stratum 1
 
 When the playbook has finished your Stratum 1 should be ready. In order to test your Stratum 1, even
-without a client installed, you can use curl.
+without a client installed, you can use `curl`.
 
 ```bash
 curl --head http://url-to-your-stratum1/cvmfs/pilot.eessi-hpc.org/.cvmfspublished
--- a/docs/filesystem_layer/stratum1.md
+++ b/docs/filesystem_layer/stratum1.md
@@ -144,10 +144,7 @@ If you already had configured the client before, you can simply reload the confi
 sudo cvmfs_config reload -c pilot.eessi-hpc.org
 ```
 
-Finally you [`verify`](https://github.com/EESSI/filesystem-layer#client) that
-the client can connect to your Stratum 1.
-
-To check which Stratum 1 you are connected to, run this command:
+Finally, verify that the client connects to your new Stratum 1 by running:
 
 ```bash
 cvmfs_config stat -v pilot.eessi-hpc.org
--- a/docs/filesystem_layer/stratum1.md
+++ b/docs/filesystem_layer/stratum1.md
@@ -126,7 +126,7 @@ curl --head http://bgo-no.stratum1.cvmfs.eessi-infra.org/cvmfs/pilot.eessi-hpc.o
 You can also test access to your Stratum 1 from a client, for which you will have to install the CVMFS
 [client](https://github.com/EESSI/filesystem-layer#clients). 
 
-Then you run this commands to add you newly created Stratum 1:
+Then run the following command to add your newly created Stratum 1 to the existing list of EESSI Stratum 1 servers by creating a local CVMFS configuration file:
 
 ```bash
 echo 'CVMFS_SERVER_URL="http://<S1_IP>/cvmfs/@fqrn@;$CVMFS_SERVER_URL"' | sudo tee -a /etc/cvmfs/domain.d/eessi-hpc.org.local
--- a/docs/filesystem_layer/stratum1.md
+++ b/docs/filesystem_layer/stratum1.md
@@ -123,10 +123,8 @@ Example with the Norwegian Stratum 1:
 curl --head http://bgo-no.stratum1.cvmfs.eessi-infra.org/cvmfs/pilot.eessi-hpc.org/.cvmfspublished
 ```
 
-You can also test access to your Stratum 1 from a client
-
-To test it you will have to install the CMVFS
-[`client`](https://github.com/EESSI/filesystem-layer#clients).
+You can also test access to your Stratum 1 from a client, for which you will have to install the CVMFS
+[client](https://github.com/EESSI/filesystem-layer#clients). 
 
 Then you run this commands to add you newly created Stratum 1:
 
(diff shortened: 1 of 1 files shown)


docs/filesystem_layer/stratum1.md
```diff
@@ -50,8 +50,16 @@ Then install Ansible roles for EESSI:
 ansible-galaxy role install -r requirements.yml -p ./roles --force
 ```
 
+Make sure you have enough space in `/srv` (on the Stratum 1) since the snapshot of the Stratum 0
+will end up there by default. If you want to alter the path where the snapshot gets copied to you
+can edit this variable in `roles/galaxyproject.cvmfs/defaults/main.yml`:
```
bedroge, line 55: Better not to edit the files from that role, but you can override this variable per host in the `inventory/host_vars` folder (or for all your hosts in `inventory/local_site_specific_vars.yml`, or per group in `inventory/group_vars`, but I guess you often want to do this per host).

docs/filesystem_layer/stratum1.md
```diff
@@ -50,8 +50,16 @@ Then install Ansible roles for EESSI:
...
+
+```bash
+cvmfs_srv_mount: /srv
+```
+
 Make sure that you have added the hostname or IP address of your server to the
-`hosts` file. Finally install the Stratum 1 using one of the two following options.
+`/inventory/hosts` file. Finally install the Stratum 1 using one of the two following options.
```
bedroge, line 62:
```suggestion
`inventory/hosts` file. Finally, install the Stratum 1 using one of the two following options.
```

docs/filesystem_layer/stratum1.md
```diff
@@ -87,14 +95,82 @@ You can request a firewall exception rule to be added for your Stratum 1 server
 
 Make sure to include the IP address of your server.
 
+## Step 3: Verification of the Stratum 1
+
+When the playbook has finished your Stratum 1 should be ready. In order to test your Stratum 1, even
+without a client installed, you can use curl.
```
bedroge, line 101:
```suggestion
without a client installed, you can use `curl`.
```

docs/filesystem_layer/stratum1.md
```diff
@@ -87,14 +95,82 @@ You can request a firewall exception rule to be added for your Stratum 1 server
...
+
+```bash
+curl --head http://url-to-your-stratum1/cvmfs/pilot.eessi-hpc.org/.cvmfspublished
+```
+This should return:
+
+```bash
+HTTP/1.1 200 OK
+...
+X-Cache: MISS from url-to-your-proxy
```
bedroge, line 111: This is actually the stratum 1's Squid proxy (might be different when you pass `--proxy <your-proxy>` to `curl` as well), so maybe we should use `url-to-your-stratum1` instead.

docs/filesystem_layer/stratum1.md
```diff
@@ -87,14 +95,82 @@ You can request a firewall exception rule to be added for your Stratum 1 server
...
+```
+
+The second time you run it, you should get a cache hit:
+
+```bash
+X-Cache: HIT from url-to-your-proxy
```
bedroge, line 117: Same here

docs/filesystem_layer/stratum1.md
```diff
@@ -87,14 +95,82 @@ You can request a firewall exception rule to be added for your Stratum 1 server
...
+Example with the Norwegian Stratum 1:
+
+```bash
+curl --head http://bgo-no.stratum1.cvmfs.eessi-infra.org/cvmfs/pilot.eessi-hpc.org/.cvmfspublished
+```
+
+You can also test access to your Stratum 1 from a client
+
+To test it you will have to install the CMVFS
+[`client`](https://github.com/EESSI/filesystem-layer#clients).
```
bedroge, line 129:
```suggestion
You can also test access to your Stratum 1 from a client, for which you will have to install the CVMFS
[client](https://github.com/EESSI/filesystem-layer#clients). 
```

bedroge: This should be moved to the documentation as well, but for now this link is okay. We can change that later on.

docs/filesystem_layer/stratum1.md
```diff
@@ -87,14 +95,82 @@ You can request a firewall exception rule to be added for your Stratum 1 server
...
+
+Then you run this commands to add you newly created Stratum 1:
```
bedroge, line 131:
```suggestion
Then run the following command to add your newly created Stratum 1 to the existing list of EESSI Stratum 1 servers by creating a local CVMFS configuration file:
```

docs/filesystem_layer/stratum1.md
```diff
@@ -87,14 +95,82 @@ You can request a firewall exception rule to be added for your Stratum 1 server
...
+echo 'CVMFS_SERVER_URL="http://<S1_IP>/cvmfs/@fqrn@;$CVMFS_SERVER_URL"' | sudo tee -a /etc/cvmfs/domain.d/eessi-hpc.org.local
+```
+
+If this is the first time you set up the client you now run:
+
+```bash
+sudo cvmfs_config setup
+```
+
+If you already have configured a client from before you reload the config:
```
bedroge, line 143:
```suggestion
If you already had configured the client before, you can simply reload the config:
```

docs/filesystem_layer/stratum1.md
```diff
@@ -87,14 +95,82 @@ You can request a firewall exception rule to be added for your Stratum 1 server
...
+
+```bash
+sudo cvmfs_config reload -c pilot.eessi-hpc.org
+```
+
+Finally you [`verify`](https://github.com/EESSI/filesystem-layer#client) that
+the client can connect to your Stratum 1.
+
+To check which Stratum 1 you are connected to, run this command:
```
bedroge, line 152:
```suggestion
Finally, verify that the client connects to your new Stratum 1 by running:
```

docs/filesystem_layer/stratum1.md
```diff
@@ -87,14 +95,82 @@ You can request a firewall exception rule to be added for your Stratum 1 server
...
+
+```bash
+cvmfs_config stat -v pilot.eessi-hpc.org
+```
+
+This should return:
```
bedroge, line 158:
```suggestion
Assuming that your new Stratum 1 is the geographically closest one to your client, this should return:
```

bedroge (changes requested):

docs/filesystem_layer/stratum1.md
rungitta: Thanks for clarifying!

docs/filesystem_layer/stratum1.md
rungitta: I thought about that. We both use `url-to-your-stratum1` and `<S1_IP_>`. Should be maybe keep to one? or somehow catch both..

bedroge: Ah, yes, better to keep that consistent across the page.

rungitta: is `url-or-ip-to-your-stratum1` too much?

bedroge: Okay with me :-)

docs/filesystem_layer/stratum1.md
```diff
@@ -50,8 +50,16 @@ Then install Ansible roles for EESSI:
 ansible-galaxy role install -r requirements.yml -p ./roles --force
 ```
 
+Make sure you have enough space in `/srv` (on the Stratum 1) since the snapshot of the Stratum 0
+will end up there by default. To alter the directory where the snapshot gets copied to you can add
+this variable in `inventory/host_vars/stratum1host`:
```
bedroge, line 55: The name of the file should be the same as the host/ip that you use in your hosts file (otherwise Ansible won't pick it up), so maybe it's better to use `<url-or-ip-to-your-stratum1>` here too:
```suggestion
this variable in `inventory/host_vars/<url-or-ip-to-your-stratum1>`:
```

bedroge (approved):
Looks good to me, thanks a lot @rungitta!

PR #86, opened 2022-03-21
Update pilot docs for 2021.12

bedroge:
Note that I've kept ReFrame (3.9.1) in the list of available tools, even though it's in the compat layer and not available as module.


boegel (approved):
lgtm

PR #98, opened 2022-12-07
Harden CI security

ocaisa:


.github/workflows/deploy.yml
```diff
@@ -3,15 +3,21 @@ name: deploy documentation (only on push to main branch)
...
     branches: main
+# Declare default permissions as read only.
+permissions: read-all
 jobs:
   build:
     runs-on: ubuntu-22.04
+    permissions:
+      contents: read
+      # Need to be able to write to the deploy branch
+      contents: write
```
boegel, line 14: This line overrules the `read` above, no?

ocaisa: Well, write implies read I think. But I don't really have any way of checking what works until we (try to) deploy.

PR #117, opened 2023-10-20
Interactive building for debugging

casparvl:
This PR moves the 'adding software' pages to a new top level header, as I was told by an external user he hadn't found it initially.

Also, added a page on how to interactively replicated the build process.


mkdocs.yml
```diff
@@ -41,6 +40,13 @@ nav:
     - using_eessi/setting_up_environment.md
     - using_eessi/basic_commands.md
     - using_eessi/eessi_demos.md
+  - Contributing software to EESSI:
```
trz42, line 43: "Contributing" is maybe not the best wording? How about "Adding ..."? Then the first item should be renamed.

It could be something like
- Adding software to EESSI
  - Overview
  - Contribution policy (requires #108)
  - Opening a pull request (anyone)
  - Building software (maintainers)
  - Debugging build issues (anyone)
  - Deploying built packages (maintainers)

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,185 @@
...
+
+For more info on starting the EESSI software environment, see [here](../using_eessi/setting_up_environment.md)
+
+### Configure EasyBuild
+It is important that we configure EasyBuild in the same way as the bot uses it, with two small exceptions:
+
+- Our working directory will be different
+- Our installpath will be different
+
+For both, any writeable path will do. In this example, we will choose `/tmp/easybuild` as our workdir, and `$HOME/.local/easybuild` as our installpath. Finally, we will source the `configure_easybuild` script, which will configure EasyBuild by setting environment variables.
```
casparvl, line 121: Take out `$HOME`

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,185 @@
...
+
+```
+export WORKDIR=/tmp/easybuild
```
casparvl, line 124: use `mktemp` instead to keep it unique, to avoid clashes between multiple users on the node

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,185 @@
...
+source configure_easybuild
+export EASYBUILD_INSTALLPATH=/tmp/easybuild
```
casparvl, line 126: Use WORKDIR so that you land in the same tempdir

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,180 @@
...
+${EESSI_CVMFS_REPO}/versions/${EESSI_PILOT_VERSION}/compat/${EESSI_OS_TYPE}/${EESSI_CPU_FAMILY}/startprefix
+```
+
+if you are unsure, you can start the EESSI software environment (see next step) and check the values of `EESSI_OS_TYPE` and `EESSI_CPU_FAMILY` set by that initialization script. Note that you'll have to start over with a new shell (i.e. quit the container) and repeat the current step of starting the Gentoo Prefix environment, as the order of those two steps matters.
+
+Now, reset the `${EESSI_CVMFS_REPO}` and `${EESSI_PILOT_VERSION}` in your prefix environment
+```
+export EESSI_CVMFS_REPO=...
+export EESSI_PILOT_VERSION=...
+```
```
trz42, line 95: Why this? Ah, because `startprefix` clears those environment variables?

Then, `run_in_compat_layer_env.sh` is the way to go, because it preserves these in some way.

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,180 @@
...
+
+!!! Note
+    By activating the Gentoo Prefix environment, the system tools (e.g. `ls`) you would normally use are now provided by Gentoo Prefix, instead of the container OS. E.g. running `which ls` after starting the prefix environment as above will return `/cvmfs/pilot.eessi-hpc.org/versions/2023.06/compat/linux/x86_64/bin/ls`. This makes the builds completely independent from the container OS.
```
trz42, line 98: Hmm, do we need to state this here?

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,180 @@
...
+!!! Note
+    If you want to replicate a build with `generic` optimization (i.e. in `$EESSI_CVMFS_REPO/versions/${EESSI_PILOT_VERSION}/software/${EESSI_OS_TYPE}/${EESSI_CPU_FAMILY}/generic`) you will need to set `export EESSI_SOFTWARE_SUBDIR_OVERRIDE=${EESSI_CPU_FAMILY}/generic` before starting the EESSI environment.
+
+To activate the software environment, run
+```
+source ${EESSI_CVMFS_REPO}/versions/${EESSI_PILOT_VERSION}/init/bash
+```
+
+!!! Note
+    If you get an error `bash: /versions//init/bash: No such file or directory`, you forgot to reset the `${EESSI_CVFMS_REPO}` and `${EESSI_PILOT_VERSION}` environment variables at the end of the previous step.
```
trz42, line 110: Possibly this could be removed if you run with `run_in_compat_layer_env.sh`

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,180 @@
...
+
+
+For more info on starting the EESSI software environment, see [here](../using_eessi/setting_up_environment.md)
+
+### Configure EasyBuild
+It is important that we configure EasyBuild in the same way as the bot uses it, with two small exceptions:
+
+- Our working directory will be different
+- Our installpath will be different
```
trz42, line 119: Unless one wants to build into `/cvmfs/...`

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,180 @@
...
+
+For both, any writeable path will do. In this example, we will choose `/tmp/easybuild` as our workdir, and `$HOME/.local/easybuild` as our installpath. Finally, we will source the `configure_easybuild` script, which will configure EasyBuild by setting environment variables.
```
trz42, line 121: Should we prefer local disks for the working directory, so avoid parallel file systems? Unsure about that one. I'd say recommend what you do when using EasyBuild locally.

trz42: Ah, one thing, if you use something under `/tmp` it may be saved into the tarball created by the `eessi_container.sh` script. If so that might be beneficial.

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Building the software
+When the bot builds software, it loops over all EasyStack files that have been changed, and builds them using EasyBuild. However, a single PR may add multiple items to a single EasyStack file, and the issue you are trying to debug is probably in _one_ of them. Getting EasyBuild to build the full EasyStack file will create the most similar situation to what the bot does. However, you _may_ just want to build the individual software that has changed. Below, we describe both approaches.
+
+### Building everything in the EasyStack file
+In our [example PR](https://github.com/EESSI/software-layer/pull/360), the EasyStack file that was changed was `eessi-2023.06-eb-4.8.1-2021b.yml`. To build this, we run (in the directory that contains the checkout of this feature branch):
+```
+eb --easystack eessi-2023.06-eb-4.8.1-2021b.yml --robot
+```
+After some time, this build fails whil trying to build `Plumed`, and we can access the build log to look for clues on why it failed.
```
trz42, line 178:
```suggestion
After some time, this build fails while trying to build `Plumed`, and we can access the build log to look for clues on why it failed.
```

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Building an individual package
+In our [example PR](https://github.com/EESSI/software-layer/pull/360), the individual package that was added to `eessi-2023.06-eb-4.8.1-2021b.yml` was `LAMMPS-23Jun2022-foss-2021b-kokkos.eb`. We'll also have to mind any options that are listed in the EasyStack file for `LAMMPS-23Jun2022-foss-2021b-kokkos.eb`, in this case the option `--from-pr 19000`. Thus, to build, we run:
```
trz42, line 181: Maybe?
```suggestion
In our [example PR](https://github.com/EESSI/software-layer/pull/360), the individual package that was added to `eessi-2023.06-eb-4.8.1-2021b.yml` was `LAMMPS-23Jun2022-foss-2021b-kokkos.eb`. We'll also have to (re)use any options that are listed in the EasyStack file for `LAMMPS-23Jun2022-foss-2021b-kokkos.eb`, in this case the option `--from-pr 19000`. Thus, to build, we run:
```

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,180 @@
...
+```
+eb LAMMPS-23Jun2022-foss-2021b-kokkos.eb --robot --from-pr 19000
+```
+After some time, this build fails whil trying to build `Plumed`, and we can access the build log to look for clues on why it failed.
```
trz42, line 185:
```suggestion
After some time, this build fails while trying to build `Plumed`, and we can access the build log to look for clues on why it failed.
```

mkdocs.yml
casparvl: One thing I'm wondering: from 'Adding software to EESSI', is it immediately clear that this is about _anyone_ being able to do that? From that header, it could also be a page that just describes how _we_ add software to EESSI.

The nice thing about the word `contributing` is that it is immediately clear that any contributor can do that.

I do like your other headers, and the suggestion to include the contribution policy under this new header.

casparvl: Never mind, I'm bikeshedding. Adding is fine, I'll do that :D

I've created this structure, with 'coming soon' pages for the things we don't have content for yet. At least the full structure is in place like this.

docs/contributing_sw/debugging_failed_builds.md
casparvl: Done

docs/contributing_sw/debugging_failed_builds.md
casparvl: Done

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,180 @@
+# Debugging failed builds
+
+Unfortunately, software does not always build successfully. Since EESSI targets novel CPU architectures as well, build failures on such platforms are quite common, as the software and/or the software build systems have not always been adjusted to support these architectures yet.
+
+In EESSI, the build are performed by a bot. This is great for builds that complete successfully as we can build a lot of software, for a wide range of hardware because of this automation. However, it does means that you, as contributor, can not easily access the build directory and build logs to figure out build issues.
```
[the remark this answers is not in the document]
casparvl: Rephrased.

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,180 @@
...
+
+This page describes how you can interactively reproduce failed builds, so that you can more easily debug the issue.
+
+Throughout this page, we will use [this PR](https://github.com/EESSI/software-layer/pull/360) as an example. It builds LAMMPS, and failed (among other things) on a [build issue for Plumed](https://github.com/EESSI/software-layer/pull/360#issuecomment-1765913105).
```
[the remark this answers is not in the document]
casparvl: Rephrased

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Starting a shell in the EESSI container
+Simply run the EESSI container (`eessi_container.sh`), which should be in the root of the `software-layer` repository
+```
+./eessi_container.sh
+```
+!!! Note
+    You may have to press enter to clearly see the prompt as some messages
+        beginning with `CernVM-FS: ` have been printed after the first prompt
+            `Apptainer> ` was shown.
```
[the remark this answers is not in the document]
casparvl: Changed

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,180 @@
...
+
+If you want to debug an issue for which a lot of dependencies need to be build first, you may want to start the container with the `--save DIR/TGZ` and flag (check `./eessi_container.sh --help`) in order to be able to resume later. E.g.
```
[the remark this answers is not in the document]
casparvl: Ok, but since we later on suggest to put our `WORKDIR` and `EASYBUILD_INSTALLPATH` in `/tmp`, I think it makes sense to keep this instruction. I've elaborated a bit in that it is specifically meant for retaining the `/tmp` data, which is the direction we later advise as workdir & installdir. Hopefully people can then make the connection themselves that if they use a different `WORKDIR` and `EASYBUILD_INSTALLDIR`, that they don't need to use `--save`.

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,180 @@
...
+
+```
+./eessi_container.sh --save ${HOME}/pr370/
```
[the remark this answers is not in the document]
casparvl: I'll add the remark that the tarballs can be sizeable, and that they should pick an FS where they have enough quotum.

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,180 @@
...
+logout
+Leaving Gentoo Prefix with exit status 1
+Apptainer> exit
+exit
+Saved contents of tmp directory '/tmp/eessi.VgLf1v9gf0' to tarball '${HOME}/pr370/EESSI-pilot-1698056784.tgz' (to resume session add '--resume ${HOME}/pr370//EESSI-pilot-1698056784.tgz')
+```
+
+Next time you want to continue investigating this issue, you can start the container with `--resume DIR/TGZ` and continue where you left off, having all dependencies already built and available.
+```
+./eessi_container.sh --resume ${HOME}/pr370//EESSI-pilot-1698056784.tgz
```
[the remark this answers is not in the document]
casparvl: fixed.

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,180 @@
...
+Before we start, check the current values of `${EESSI_CVMFS_REPO}` and `${EESSI_PILOT_VERSION}` so that you can reset them later:
+```
+echo ${EESSI_CVMFS_REPO}
+echo ${EESSI_PILOT_VERSION}
+```
+
+To do that, you need to run the `startprefix` command. However, we have several compatibility layers, and you'll need to run it for the one that matches the host node. For example, on an aarch64 (ARM) linux machine:
+```
+export EESSI_OS_TYPE=linux
+export EESSI_CPU_FAMILY=aarch64
```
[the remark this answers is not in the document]
casparvl: Hm, so I now used the `minimal_eessi_env`  script or whatever its called. But you're right, we can probably just extract the two lines that are needed from this script. That makes it more transparent what's happening, and also less dependency on this script to stay around.

docs/contributing_sw/debugging_failed_builds.md
casparvl: We discussed this, it's not an option since `run_in_compat_layer_env.sh` has no option of dropping you in an interactive shell - it runs some commands and then returns to a shell outside of prefix.

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,180 @@
...
+${EESSI_CVMFS_REPO}/versions/${EESSI_PILOT_VERSION}/compat/${EESSI_OS_TYPE}/${EESSI_CPU_FAMILY}/startprefix
+```
+
+if you are unsure, you can start the EESSI software environment (see next step) and check the values of `EESSI_OS_TYPE` and `EESSI_CPU_FAMILY` set by that initialization script. Note that you'll have to start over with a new shell (i.e. quit the container) and repeat the current step of starting the Gentoo Prefix environment, as the order of those two steps matters.
```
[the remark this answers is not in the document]
casparvl: Removed it, as we now set it using `uname -m` (and linux, hardcoded).

docs/contributing_sw/debugging_failed_builds.md
casparvl: Well, it gives some idea of _why_ we do this, and also provides a way for users who are in doubt to easily check if they are now in a prefix environment. I think as a note, it's ok.

docs/contributing_sw/debugging_failed_builds.md
casparvl: See above

docs/contributing_sw/debugging_failed_builds.md
casparvl: Agreed. I've had one issue that didn't show up when building somewhere else. But: I don't know what needs to change exactly to build in `/cvmfs/`. I guess we need to pass `--access rw` then? Does it then mount the writeable overlay? Or is there more that needs to be done?

As I understand, it is a bit slower to compile in the writeable overlay. What would be your idea: should we advise that by default, simply to stay as close to the bot as possible? Or not, because it is slower?

docs/contributing_sw/debugging_failed_builds.md
casparvl: Yeah, it is. I've tried that combined with `--save`. Indeed, I could easily resume, having some dependencies already build in `/tmp` the first time. Saved me time :)

docs/contributing_sw/debugging_failed_builds.md
casparvl: Also, if we install there, is that saved into the tarball as well by `--save` and then restored by `--restore`? I.e. if I need to build some deps first,, then save, then restore later, will the deps be already there?

docs/contributing_sw/debugging_failed_builds.md
casparvl: Done

docs/contributing_sw/debugging_failed_builds.md
casparvl: Done

docs/contributing_sw/debugging_failed_builds.md
casparvl: Done

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,189 @@
...
+When the bot builds software, it loops over all EasyStack files that have been changed, and builds them using EasyBuild. However, a single PR may add multiple items to a single EasyStack file, and the issue you are trying to debug is probably in _one_ of them. Getting EasyBuild to build the full EasyStack file will create the most similar situation to what the bot does. However, you _may_ just want to build the individual software that has changed. Below, we describe both approaches.
+
+### Building everything in the EasyStack file
+In our [example PR](https://github.com/EESSI/software-layer/pull/360), the EasyStack file that was changed was `eessi-2023.06-eb-4.8.1-2021b.yml`. To build this, we run (in the directory that contains the checkout of this feature branch):
+```
+eb --easystack eessi-2023.06-eb-4.8.1-2021b.yml --robot
+```
+After some time, this build fails while trying to build `Plumed`, and we can access the build log to look for clues on why it failed.
+
+### Building an individual package
```
trz42, line 184: Maybe one should caution here, that the bot is "never" running `eb` to install packages like that. So, the command could just work and not fail simply because it is just a little different.

We hit such false positives when debugging failing builds of `PROJ-9.0.0-GCCcore-11.3.0.eb`. The error message in the bot's EB build log was not telling much, and instead of processing the easystack file we just ran the command to install a single `.eb` file in our debugging session.

trz42 (changes requested):
Two more suggestions.

I think we should wait merging this PR until #108 has been merged and a short overview (`overview.md`) is added.

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,189 @@
...
+### Configure EasyBuild
+It is important that we configure EasyBuild in the same way as the bot uses it, with two small exceptions:
+
+- Our working directory will be different
+- Our installpath will be different
+
+For both, any writeable path will do. In this example, we create a unique temporary directory inside `/tmp` to serve both as our workdir and installpath. Finally, we will source the `configure_easybuild` script, which will configure EasyBuild by setting environment variables.
+
+```
+export WORKDIR=$(mktemp --directory --tmpdir=/tmp  -t eessi.XXXXXXXXXX)
```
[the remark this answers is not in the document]
casparvl: I'll make it `eessi-debug.XXXXXXX`. It's still good to know it is related to EESSI, in case you find a stray dir like this in your `/tmp`.

docs/contributing_sw/debugging_failed_builds.md
casparvl: Done, put a note with this caveat.

docs/contributing_sw/debugging_failed_builds.md
casparvl: Ok, `--access rw` and _not_ setting `EASYBUILD_INSTALLPATH` ourselves (i.e. relying on the one set by `configure_easybuild` works. This creates builds that actually get installed into the writeable overlay. That has the advantage of being even more similar to what the bot does. Hence, I was able to replicate also [this issue](https://github.com/EESSI/software-layer/pull/332#issuecomment-1775374260).

I think that is an improvement. I still add to the instructions the suggestion of setting an `EASYBUILD_INSTALLPATH`, but only _after_ having replicated the issue - as a way of identifying if the writeable overlay is somehow causing an issue.

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,196 @@
...
+```
+
+Then, we set `EESSI_OS_TYPE` and `EESSI_CPU_FAMILY` and run the `startprefix` command to start the Gentoo Prefix environment:
+```
+export EESSI_OS_TYPE=linux  # We only support Linux for now
+export EESSI_CPU_FAMILY=$(uname -m)
+${EESSI_CVMFS_REPO}/versions/${EESSI_PILOT_VERSION}/compat/${EESSI_OS_TYPE}/${EESSI_CPU_FAMILY}/startprefix
+```
+
+Now, reset the `${EESSI_CVMFS_REPO}` and `${EESSI_PILOT_VERSION}` in your prefix environment
```
t1mk1k, line 92:
```suggestion
Now, reset the `${EESSI_CVMFS_REPO}` and `${EESSI_PILOT_VERSION}` in your prefix environment with the initial values (printed in the echo statements above)
```

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,196 @@
...
+    If you want to replicate a build with `generic` optimization (i.e. in `$EESSI_CVMFS_REPO/versions/${EESSI_PILOT_VERSION}/software/${EESSI_OS_TYPE}/${EESSI_CPU_FAMILY}/generic`) you will need to set `export EESSI_SOFTWARE_SUBDIR_OVERRIDE=${EESSI_CPU_FAMILY}/generic` before starting the EESSI environment.
+
+To activate the software environment, run
+```
+source ${EESSI_CVMFS_REPO}/versions/${EESSI_PILOT_VERSION}/init/bash
+```
+
+!!! Note
+    If you get an error `bash: /versions//init/bash: No such file or directory`, you forgot to reset the `${EESSI_CVFMS_REPO}` and `${EESSI_PILOT_VERSION}` environment variables at the end of the previous step.
+!!! Note
```
t1mk1k, line 112:
```suggestion

!!! Note
```

Extra line so notes are not on the same line

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,196 @@
...
+In this example, we create a unique temporary directory inside `/tmp` to serve both as our workdir. Finally, we will source the `configure_easybuild` script, which will configure EasyBuild by setting environment variables.
+
+```
+export WORKDIR=$(mktemp --directory --tmpdir=/tmp  -t eessi-debug.XXXXXXXXXX)
+source configure_easybuild
+```
+Among other things, the `configure_easybuild` script sets the install path for EasyBuild to point to the correct installation directory in (to `${EESSI_CVMFS_REPO}/versions/${EESSI_PILOT_VERSION}/software/${EESSI_OS_TYPE}/${EESSI_SOFTWARE_SUBDIR}`). This is the exact same path the `bot` uses to build, and uses a writeable overlay filesystem in the container to write to a path in `/cvmfs` (which normally is read-only). Since this is identical to what the `bot` does, we advise you to start with this when reproducting a build failure. However, after having reproduced the bug, you may want to set a different `EASYBUILD_INSTALLPATH`, e.g.
+
+```
+export EASYBUILD_INSTALLPATH=${WORKDIR}
```
t1mk1k, line 130: Might be good to add this step after going describing the initial debug run. To us copy-paste monkeys, this one looks like one of the commands you should use, if you skim (or fail to read) the text above.

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,196 @@
...
+```
+
+(_after_ sourcing the `configure_easybuild` script, so that you overwrite whatever that script sets). This can help you identify if an issue is related to building in a writeable overlay. For example, the writeable overlay is know to be a bit slow sometimes, and we have seen tests failing because they exceeded some timeout.
```
t1mk1k, line 133:
```suggestion
(_after_ sourcing the `configure_easybuild` script, so that you overwrite whatever that script sets). This can help you identify if an issue is related to building in a writeable overlay. For example, the writeable overlay is known to be a bit slow sometimes, and we have seen tests failing because they exceeded some timeout.
```

docs/contributing_sw/debugging_failed_builds.md
```diff
@@ -0,0 +1,196 @@
...
+
+!!! Note
+    If you started the container using --resume, yoy may want WORKDIR to point to the workdir you created previously (instead of create a new, temporary directory with `mktemp`).
```
t1mk1k, line 136:
```suggestion
    If you started the container using --resume, you may want WORKDIR to point to the workdir you created previously (instead of creating a new, temporary directory with `mktemp`).
```

docs/contributing_sw/debugging_failed_builds.md
casparvl: Changed

docs/contributing_sw/debugging_failed_builds.md
casparvl: Changed

docs/contributing_sw/debugging_failed_builds.md
casparvl: Agreed, I made a new section on known causes of issues (after the 'build' instructions), and moved it there.

docs/contributing_sw/debugging_failed_builds.md
casparvl: Rephrased this whole section, so this typo is gone now.

docs/contributing_sw/debugging_failed_builds.md
casparvl: fixed

trz42 (approved):

trz42:
Looks good to me. Thanks @casparvl !

PR #118, opened 2023-10-20
Added documentation on how to build on top of EESSI

casparvl:
As discussed on Slack, there is much we can do to make this easier with a module. Still, I think it's useful to have _some_ documentation in place as soon as we can. We can then take it from there if there are improvements :)

### DIFF
--- a/docs/using_eessi/building_on_eessi.md
+++ b/docs/using_eessi/building_on_eessi.md
@@ -1,7 +1,7 @@
 # Building software on top of EESSI
 
 ## Building software on top of EESSI with EasyBuild
-Building on top of EESSI with EasyBuild is relatively straightforward. One crucial feature is that EasyBuild supports building against operating system libraries that are _not_ in a standard prefix (such as `/usr/lib`). This is required when building against EESSI, since all of the software in EESSI is build against the [compatibility layer](../compatibility_layer.md).
+Building on top of EESSI with EasyBuild is relatively straightforward. One crucial feature is that EasyBuild supports building against operating system libraries that are _not_ in a standard prefix (such as `/usr/lib`). This is required when building against EESSI, since all of the software in EESSI is built against the [compatibility layer](../compatibility_layer.md).
 
 ### Starting the EESSI software environment
 Start your environment as described [here](../using_eessi/setting_up_environment.md)
@@ -46,7 +46,7 @@ eb netCDF-4.9.0-gompi-2022a.eb
 ```
 
 !!! Note
-    If this netCDF module is available by the time you are trying, you can force a local rebuild by adding the `--force` argument in order to experiment with building locally, or pick a different EasyConfig to build.
+    If this netCDF module is available by the time you are trying, you can force a local rebuild by adding the `--rebuild` argument in order to experiment with building locally, or pick a different EasyConfig to build.
 
 ### Using the newly built module
 First, you'll need to add the subdirectory of the `EASYBUILD_INSTALLPATH` that contains the modules to the `MODULEPATH`. You can do that using:
@@ -60,7 +60,7 @@ you may want to do this as part of your `.bashrc`.
 !!! Note
     Be careful adding to the `MODULEPATH` in your `.bashrc` if you are on a cluster with heterogeneous architectures. You don't want to pick up on a module that was not compiled for the correct architectures accidentally.
 
-Since your module is build on top of the EESSI environment, that needs to be loaded first (as described [here](../using_eessi/setting_up_environment.md)), if you haven't already done so.
+Since your module is built on top of the EESSI environment, that needs to be loaded first (as described [here](../using_eessi/setting_up_environment.md)), if you haven't already done so.
 
 Finally, you should be able to load our newly build module:
 ```
@@ -70,4 +70,4 @@ module load netCDF/4.9.0-gompi-2022a
 ## Manually building software op top of EESSI
 Building software on top of EESSI would require your linker to use the same system-dependencies as the software in EESSI does. In other words: it requires you to link against libraries from the compatibility layer, instead of from your host OS.
 
-While we plan to support this in the future, manually building on top of EESSI is not supported yet.
+While we plan to support this in the future, manually building on top of EESSI is currently not supported yet in a trivial way.
--- a/docs/using_eessi/building_on_eessi.md
+++ b/docs/using_eessi/building_on_eessi.md
@@ -24,8 +24,8 @@ cd software-layer
 Then, you have to pick a working directory (that you have write access to) where EasyBuild can do the build, and an install directory (with sufficient storage space), where EasyBuild can install it. In this example, we create a temporary directory in `/tmp/` as our working directory, and use `$HOME/.local/easybuild` as our installpath:
 ```
 export WORKDIR=$(mktemp --directory --tmpdir=/tmp  -t eessi-build.XXXXXXXXXX)
-export EASYBUILD_INSTALLPATH="${HOME}/.local/easybuild"
 source configure_easybuild
+export EASYBUILD_INSTALLPATH="${HOME}/.local/easybuild"
 ```
 Next, you load the EasyBuild module that you want to use, e.g. 
 ```
--- a/docs/using_eessi/building_on_eessi.md
+++ b/docs/using_eessi/building_on_eessi.md
@@ -29,7 +29,7 @@ export EASYBUILD_INSTALLPATH="${HOME}/.local/easybuild"
 ```
 Next, you load the EasyBuild module that you want to use, e.g. 
 ```
-module load EasyBuild/4.8.1
+module load EasyBuild/4.8.2
 ```
 Finally, you can check the current configuration for EasyBuild using
 ```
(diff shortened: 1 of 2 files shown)


docs/using_eessi/building_on_eessi.md
```diff
@@ -0,0 +1,69 @@
+# Building software on top of EESSI
+
+## Building software on top of EESSI with EasyBuild
+Building on top of EESSI with EasyBuild is relatively straightforward. One crucial feature is that EasyBuild supports building against operating system libraries that are _not_ in a standard prefix (such as `/usr/lib`). This is required when building against EESSI, since all of the software in EESSI is build against the [compatibility layer](../compatibility_layer.md).
+
+### Starting the EESSI software environment
+Start your environment as described [here](../using_eessi/setting_up_environment.md)
+
+### Configure EasyBuild
+To configure EasyBuild, first, check out the Github repository. We advice you check out the branch corresponding to the version of EESSI you would like to use.
```
trz42, line 10: Maybe mention the software-layer repo already here?
```suggestion
To configure EasyBuild, first, check out the [EESSI software-layer repository](https://github.com/EESSI/software-layer.git). We advise you to check out the branch corresponding to the version of EESSI you would like to use.
```

docs/using_eessi/building_on_eessi.md
```diff
@@ -0,0 +1,69 @@
...
+
+To build on top of e.g. version `2023.06` of the EESSI software stack, we check it out, and go into that directory:
+
+```
+git clone https://github.com/EESSI/software-layer/ --branch 2023.06
+cd software-layer
+```
+Then, you have to pick a working directory (that you have write access to) where EasyBuild can do the build, and an install directory, where EasyBuild can install it. In this example, we use `/tmp/easybuild` as our working directory, and `$HOME/.local/easybuild` as our installpath:
+```
+export WORKDIR=/tmp/easybuild
```
trz42, line 26: Should we use something similar to what is suggested in #117? For example,
```bash
export WORKDIR=$(mktemp --directory --tmpdir=/tmp  -t eessi.XXXXXXXXXX)
```
Although, `eessi` should probably be replaced with something else, e.g., `build`, to make it easier to distinguish different temp directories from each other ?

docs/using_eessi/building_on_eessi.md
```diff
@@ -0,0 +1,69 @@
...
+export EASYBUILD_INSTALLPATH="${HOME}/.local/easybuild"
```
trz42, line 27: We should caution that quota for `${HOME}` may be very small, hence not to much software should be installed in that location.

docs/using_eessi/building_on_eessi.md
```diff
@@ -0,0 +1,69 @@
...
+```
+module load EasyBuild/4.8.1
+```
+Finally, you can check the current configuration for EasyBuild using
+```
+eb --show-config
+```
+
+!!! Note
+    We use EasyBuild's default behaviour in optimizing for the host architecture. Since the EESSI initialization script also loads the EESSI stack that is optimized for your host architecture, this matches nicely. However, if you work on a cluster with heterogeneous node types, you have to realize you can only use these builds on the same architecture as where you build them. You can use different `EASYBUILD_INSTALLPATH`s if you want to build for different host architectures.
```
trz42, line 40: Hmm, if I had a `zen4` system what would happen? Would I build for `zen4` or `zen3` (or what the best fit is)?

docs/using_eessi/building_on_eessi.md
```diff
@@ -0,0 +1,69 @@
...
+Now, you are ready to build. For example, at the time of writing, `netCDF-4.9.0-gompi-2022a.eb` was not in the EESSI environment yet, so you can build it yourself:
+```
+eb netCDF-4.9.0-gompi-2022a.eb
+```
+
+!!! Note
+    If this netCDF module is available by the time you are trying, you can force a local rebuild by adding the `--force` argument in order to experiment with building locally, or pick a different EasyConfig to build.
+
+### Adding your module directory to the `MODULEPATH`
+Finally, we need to make sure this module is available on our `MODULEPATH`. The path where EasyBuild installs modules by default is `${EASYBUILD_INSTALLPATH}/modules/all`. You can add it to your modulepath by running:
```
trz42, line 52: Maybe?
```suggestion
Finally, we need to make sure this module is available on our `MODULEPATH`. The path where EasyBuild installs modules by default is `${EASYBUILD_INSTALLPATH}/modules/all`. You can add it to your `MODULEPATH` by running:
```

docs/using_eessi/building_on_eessi.md
```diff
@@ -0,0 +1,69 @@
...
+```
+module use ${EASYBUILD_INSTALLPATH}/modules/all
+```
+You may want to do this as part of your `.bashrc`.
+
+Now, we should be able to load our newly build module:
+```
+module load netCDF/4.9.0-gompi-2022a
```
trz42, line 60: Would this work in a fresh shell (without sourcing the EESSI init script first)?

trz42 (changes requested):
Overall looks good. Left a few comments.

Haven't tried it yet.

docs/using_eessi/building_on_eessi.md
casparvl: Changed

docs/using_eessi/building_on_eessi.md
casparvl: done, will call it `eessi-build.XXXXXXXXX`

docs/using_eessi/building_on_eessi.md
casparvl: I've added a note that the install directory should have sufficient storage space. However, I think it already quite clearly described that this is an example. It's also the default for EasyBuild to install (I believe), so it's not _that_ strange to give this as an example. It is also the only non-system specific directory that we can be sure is writeable to the user (even if it _may_ be system specific whether you have sufficient quotum there). I'm assuming if people don't have enough quotum, they are clever enough to put it somewhere else... :)

docs/using_eessi/building_on_eessi.md
casparvl: I'm not really sure what you are asking here. If you are building on a `zen4` host, you optimize for that, since EasyBuild's default is to optimize for the build host. 

I'll make the note more elaborate with a concrete example though.

docs/using_eessi/building_on_eessi.md
casparvl: Rephrased this whole section, now referring to it as `MODULEPATH` as well.

docs/using_eessi/building_on_eessi.md
```diff
@@ -0,0 +1,73 @@
...
+
+```
+git clone https://github.com/EESSI/software-layer/ --branch 2023.06
+cd software-layer
+```
+Then, you have to pick a working directory (that you have write access to) where EasyBuild can do the build, and an install directory (with sufficient storage space), where EasyBuild can install it. In this example, we create a temporary directory in `/tmp/` as our working directory, and use `$HOME/.local/easybuild` as our installpath:
+```
+export WORKDIR=$(mktemp --directory --tmpdir=/tmp  -t eessi-build.XXXXXXXXXX)
+export EASYBUILD_INSTALLPATH="${HOME}/.local/easybuild"
+source configure_easybuild
```
trz42, line 28: The order of lines 27 and 28 need to be swapped. That's because `configure_easybuild` sets `EASYBUILD_INSTALLPATH` to `${EESSI_PREFIX}/software/${EESSI_OS_TYPE}/${EESSI_SOFTWARE_SUBDIR}` without checking if it is already set.

```suggestion
export WORKDIR=$(mktemp --directory --tmpdir=/tmp  -t eessi-build.XXXXXXXXXX)
source configure_easybuild
export EASYBUILD_INSTALLPATH="${HOME}/.local/easybuild"
```

docs/using_eessi/building_on_eessi.md
```diff
@@ -0,0 +1,73 @@
...
+```
+Next, you load the EasyBuild module that you want to use, e.g. 
+```
+module load EasyBuild/4.8.1
```
trz42, line 32: Minor suggestion, use version 4.8.2.

docs/using_eessi/building_on_eessi.md
```diff
@@ -0,0 +1,73 @@
...
+### Building
+Now, you are ready to build. For example, at the time of writing, `netCDF-4.9.0-gompi-2022a.eb` was not in the EESSI environment yet, so you can build it yourself:
+```
+eb netCDF-4.9.0-gompi-2022a.eb
+```
+
+!!! Note
+    If this netCDF module is available by the time you are trying, you can force a local rebuild by adding the `--force` argument in order to experiment with building locally, or pick a different EasyConfig to build.
+
+### Using the newly build module
```
trz42, line 51: Probably meant _built_?

trz42 (changes requested):
Just tested the procedure. It worked after doing one important swap of commands in the configuration. Added two minor suggestions.

Should be easy to change.

ocaisa (approved):

PR #136, opened 2023-12-06
fix image urls + update to menu bar

trz42:
- image URLs require one more level up
- updated menu bar entry for overview page
- removed obsolete comments from mkdocs.yml file

### DIFF
--- a/docs/adding_software/overview.md
+++ b/docs/adding_software/overview.md
@@ -5,7 +5,7 @@ We welcome contributions to the EESSI software stack. This page shows the proced
 ## Contribute a software to the EESSI software stack
 
 <p align="center">
-  <img width="50%" src="../img/contribute_software_workflow.png" alt="EESSI contribute software">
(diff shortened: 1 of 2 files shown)


xinan1911:
> > Hi @trz42 , in the preview of the code the images are shown correctly so are the imageURLs wrong? I think because the images are png not jpg and that's why the images are not shown. What do you think?
> 
> The preview seems to be misleading in this case. I tested it locally with `mkdocs serve`, then in a browser it was looking fine.
> 
> The difference between preview and browser seems that in the preview the URL of the page is `docs/adding_software/overview.md` (note the suffix `.md` and the missing `/`), while when it is served by a webserver the URL is `docs/adding_software/overview/`. Knowing the latter it makes sense that you need `../../img/...` for the image URLs.

Thanks for the explanation. I got it. I checked the docs repository and I think this implementation https://github.com/EESSI/docs/tree/main/docs/meetings with markdown files and images are resilient to changes of folder structure.

boegel (approved):

PR #139, opened 2023-12-21
show expected and problematic output produced by deviceQuery in GPU docs

boegel:
showing output in case it doesn't work is useful for searching purposes...

### DIFF
--- a/docs/gpu.md
+++ b/docs/gpu.md
@@ -152,10 +152,32 @@ The only scenario where this would be required is if `$LD_LIBRARY_PATH` is modif
 
 ### Testing the GPU support {: #gpu_cuda_testing }
 
-The quickest way to test if software installations included in  EESSI can access and use your GPU is to run the
+The quickest way to test if software installations included in EESSI can access and use your GPU is to run the
 `deviceQuery` executable that is part of the `CUDA-Samples` module:
-```
+```{ .bash .copy }
 module load CUDA-Samples
 deviceQuery
 ```
-If both are successful, you should see information about your GPU printed to your terminal.
+If both are successful, you should see information about your GPU printed to your terminal, for example:
+
+```
+$ deviceQuery
+deviceQuery Starting...
+
+ CUDA Device Query (Runtime API) version (CUDART static linking)
+
+Detected 1 CUDA Capable device(s)
+
+Device 0: "NVIDIA A2"
+  CUDA Driver Version / Runtime Version          12.2 / 12.1
+  CUDA Capability Major/Minor version number:    8.6
+...
+```
+
+If the `deviceQuery` command can not access your GPU, you will see an error message like:
+```
+cudaGetDeviceCount returned 35
+-> CUDA driver version is insufficient for CUDA runtime version
+Result = FAIL
+```
+```


docs/gpu.md
```diff
@@ -152,10 +152,32 @@ The only scenario where this would be required is if `$LD_LIBRARY_PATH` is modif
...
+
+Detected 1 CUDA Capable device(s)
+
+Device 0: "NVIDIA A2"
+  CUDA Driver Version / Runtime Version          12.2 / 12.1
+  CUDA Capability Major/Minor version number:    8.6
+...
+```
+
+If the `deviceQuery` command can not access your GPU, you will see an error message like:
```
ocaisa, line 177: This shouldn't actually happen though, because of the Lmod guards the only scenario I can see where you would reach this is where you are using a container and the system drivers are too old

boegel: I triggered it by cleaning out the `host_injections` directory after loading the module.

I agree it's very unlikely that it happens, but we should mention it in the docs regardless, if only to let people easily find this page when searching for error messages.

ocaisa: My concern here is that the placement here makes it seem like it not working is likely, but reaching this message is actually very unlikely

ocaisa: Maybe a little box saying `What does it look like if the command fails?`

docs/gpu.md
```diff
@@ -152,10 +152,32 @@ The only scenario where this would be required is if `$LD_LIBRARY_PATH` is modif
 
 ### Testing the GPU support {: #gpu_cuda_testing }
```
casparvl, line 153: Currently, this only treats testing if you can _run_ CUDA-enabled software from EESSI. Maybe we can also include a small instruction for testing if _building new CUDA software_ on top of EESSI works properly. Something like:
```
module load CUDA/<something>
nvcc -o hello_world hello_world.cu hello_world.c (or whatever the correct syntax is here :))
./hello_world
```

casparvl: And mention they should test this for each version of CUDA they installed in `host_injections`

boegel: Makes sense, but that should be done in a separate PR?

docs/gpu.md
```diff
@@ -152,10 +152,32 @@ The only scenario where this would be required is if `$LD_LIBRARY_PATH` is modif
...
+...
+```
+
+If the `deviceQuery` command can not access your GPU, you will see an error message like:
+```
+cudaGetDeviceCount returned 35
+-> CUDA driver version is insufficient for CUDA runtime version
+Result = FAIL
+```
+```
```
ocaisa, line 183:
```suggestion
!!! note "What if the `deviceQuery` command fails?"

    If the `deviceQuery` command cannot access your GPU, you will see an error message like:
    ```
    cudaGetDeviceCount returned 35
    -> CUDA driver version is insufficient for CUDA runtime version
    Result = FAIL
    ```

```

docs/gpu.md
casparvl: If you want, sure. I won't block this one over it :) Although I would consider it to be an integral part of "Testing the GPU support" to be honest :)

ocaisa: I don't see it as so integral if we focused on software consumers, it's only integral if you want to do development-type work

PR #141, opened 2024-01-03
add new repository page for sofware.eessi.io

laraPPr:

PR #144, opened 2024-01-19
replace pilot with EESSI repo

laraPPr:
- https://github.com/EESSI/docs/pull/141


casparvl (approved):
Lgtm! I haven't checked extensively if we missed any 'pilot' reference, but if we encounter them later, we can always do a follow-up PR. At least with this, the vast majority should be gone.

PR #147, opened 2024-01-23
Set working EasyBuild version in debug docs

APN-Pucky:


boegel:
We indeed don't have EasyBuild/4.8.1 in `software.eessi.io`, but more changes should be made to this page to provide a more up-to-date example, since the easystack file has moved as well, and the EasyBuild version is tied to the easystack file being used...

laraPPr:
Thank you @APN-Pucky for reporting this we have included your suggestion in the following pr
- [ ] https://github.com/EESSI/docs/pull/497

PR #148, opened 2024-01-26
Update source command with duplicate "versions"

xinan1911:
See https://www.eessi.io/docs/using_eessi/setting_up_environment/


bedroge (approved):
Thanks @xinan1911!

PR #150, opened 2024-02-01
add OSU documentation to available tests

laraPPr:


casparvl (approved):
Lgtm! We should remember though to update the docs when Kenneth changes the test names...

PR #157, opened 2024-02-15
Update page about setting up a private Stratum 1

bedroge:
Still sort of WIP, as I need to test some commands, and need to update the filesystem-layer repo (it should use the sync server as stratum 0 there).

### DIFF
--- a/docs/filesystem_layer/stratum1.md
+++ b/docs/filesystem_layer/stratum1.md
@@ -72,12 +72,20 @@ Then install Ansible roles for EESSI:
 ansible-galaxy role install -r ./requirements.yml --force
 ```
 
-Make sure you have enough space in `/srv` on the Stratum 1, since the snapshot of the repositories
-will end up there by default. To alter the directory where the snapshots get stored you can add
-the following variable in `inventory/host_vars/<url-or-ip-to-your-stratum1>`:
+Make sure you have enough space in `/srv` on the Stratum 1, since the snapshots of the repositories
+will end up there by default. To alter the directory where the snapshots get stored you can manually
+create a symlink before running the playbook:
+```bash
+sudo ln -s /lots/of/space/cvmfs /srv/cvmfs
+```
+
+Alternatively, you can instruct the Ansible role to create a file system (e.g. ext4) on a given storage device
+and mount it at `/srv` by adding the following variable
+in `inventory/host_vars/<url-or-ip-to-your-stratum1>`:
 
 ```bash
-cvmfs_srv_mount: /lots/of/space
+cvmfs_srv_device: /dev/somedevice
+cvmfs_srv_fstype: ext4
 ```
 
 Also make sure that you have added the hostname or IP address of your server to the
--- a/docs/filesystem_layer/stratum1.md
+++ b/docs/filesystem_layer/stratum1.md
@@ -1,7 +1,7 @@
 # Setting up a Stratum 1
 
 The EESSI project provides a number of geographically distributed public Stratum 1 servers that you can use to make EESSI available on your machine(s).
-If you want to be better protected against network outages and increase the bandwith between your cluster nodes and the Stratum 1 servers,
+If you want to be better protected against network outages and increase the bandwidth between your cluster nodes and the Stratum 1 servers,
 you could consider setting up a local (private) Stratum 1 server that replicates the EESSI CVMFS repository.
 This guarantees that you always have a full and up-to-date copy of the entire stack available in your local network. 
 
--- a/docs/filesystem_layer/stratum1.md
+++ b/docs/filesystem_layer/stratum1.md
@@ -35,7 +35,7 @@ However, EESSI provides a synchronisation server that can be used for setting up
     In all cases this was due to an intrusion prevention system scanning the associated network, and hence scanning all files going in or out of the Stratum 1.
     Though it was a false-positive in all cases, this breaks the synchronization procedure of your Stratum 1.
     If this is the case, you can try switching to HTTPS by using `https://aws-eu-west-s1-sync.eessi.science` for synchronizing your Stratum 1.
-    Even though there is no advantage for CVMFS itself in using HTTPS (it has built-in mechasnims for ensuring the integrity of the data),
+    Even though there is no advantage for CVMFS itself in using HTTPS (it has built-in mechanisms for ensuring the integrity of the data),
     this will prevent the described issues, as the intrusion prevention system will not be able to inspect the encrypted data.
     As HTTPS does introduce some overhead due to the encryption/decryption, it is still recommended to use HTTP as default.
 
--- a/docs/filesystem_layer/stratum1.md
+++ b/docs/filesystem_layer/stratum1.md
@@ -24,6 +24,19 @@ Stratum 1 servers usually replicate from the Stratum 0 server.
 In order to ensure the stability and security of the EESSI Stratum 0 server, it has a strict firewall, and only the EESSI-maintained public Stratum 1 servers are allowed to replicate from it.
 However, EESSI provides a synchronisation server that can be used for setting up private Stratum 1 replica servers, and this is available at `http://aws-eu-west-s1-sync.eessi.science`.
 
+!!! warn Potential issues with intrusion prevention systems
+    In the past we have seen a few occurrences of data transfer issues when files were being pulled in by or from a Stratum 1 server.
+    In such cases the `cvmfs_server snapshot` command, used for synchronizing the Stratum 1, may break with errors like `failed to download <URL to file>`.
+    Trying to manually download the mentioned file with `curl` will also not work, and result in errors like:
+    ```
+    curl: (56) Recv failure: Connection reset by peer
+    ```
+    In all cases this was due to an intrusion prevention system scanning the associated network, and hence scanning all files going in or out of the Stratum 1.
+    Though it was a false-positive in all cases, this breaks the synchronization procedure of your Stratum 1.
+    If this is the case, you can try switching to HTTPS by using `https://aws-eu-west-s1-sync.eessi.science` for synchronizing your Stratum 1.
+    Even though there is no advantage for CVMFS itself in using HTTPS (it has built-in mechasnims for ensuring the integrity of the data),
+    this will prevent the described issues, as the intrusion prevention system will not be able to inspect the encrypted data.
+
 ### Manual configuration
 
 In order to set up a Stratum 1 manually, you can make use of the instructions in the [Private Stratum 1 replica server](https://multixscale.github.io/cvmfs-tutorial-hpc-best-practices/access/stratum1/)
--- a/docs/filesystem_layer/stratum1.md
+++ b/docs/filesystem_layer/stratum1.md
@@ -119,7 +119,7 @@ curl --head http://aws-eu-central-s1.eessi.science/cvmfs/software.eessi.io/.cvmf
 
 You can, of course, also test access to your Stratum 1 from a client.
 This requires you to install a CernVM-FS client and add the Stratum 1 to the client configuration;
-this is explained in more detail on the [native installation page](getting_access/native_installation.md).
+this is explained in more detail on the [native installation page](../getting_access/native_installation.md).
 
 Then verify that the client connects to your new Stratum 1 by running:
 
--- a/docs/filesystem_layer/stratum1.md
+++ b/docs/filesystem_layer/stratum1.md
@@ -123,7 +123,7 @@ You can also test access to your Stratum 1 from a client, for which you will hav
 Then run the following command to prepend your newly created Stratum 1 to the existing list of EESSI Stratum 1 servers by creating a local CVMFS configuration file:
 
 ```bash
-echo 'CVMFS_SERVER_URL="http://<url-or-ip-to-your-stratum1>/cvmfs/@fqrn@;$CVMFS_SERVER_URL"' | sudo tee -a /etc/cvmfs/domain.d/eessi-hpc.org.local
+echo 'CVMFS_SERVER_URL="http://<url-or-ip-to-your-stratum1>/cvmfs/@fqrn@;$CVMFS_SERVER_URL"' | sudo tee -a /etc/cvmfs/domain.d/eessi.io.local
 ```
 
 !!! note
--- a/docs/filesystem_layer/stratum1.md
+++ b/docs/filesystem_layer/stratum1.md
@@ -36,6 +36,7 @@ However, EESSI provides a synchronisation server that can be used for setting up
     If this is the case, you can try switching to HTTPS by using `https://aws-eu-west-s1-sync.eessi.science` for synchronizing your Stratum 1.
     Even though there is no advantage for CVMFS itself in using HTTPS (it has built-in mechasnims for ensuring the integrity of the data),
     this will prevent the described issues, as the intrusion prevention system will not be able to inspect the encrypted data.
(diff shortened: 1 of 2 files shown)


bedroge:
PR for changing the Stratum 0 url to the sync server: https://github.com/EESSI/filesystem-layer/pull/177.

bedroge:
This should be more or less ready now, except that one step of our `stratum1.yml` playbook currently doesn't work due to the issue that I tried to solve with https://github.com/galaxyproject/ansible-cvmfs/pull/32.

docs/filesystem_layer/stratum1.md
```diff
@@ -1,44 +1,68 @@
...
+    In the past we have seen a few occurrences of data transfer issues when files were being pulled in by or from a Stratum 1 server.
+    In such cases the `cvmfs_server snapshot` command, used for synchronizing the Stratum 1, may break with errors like `failed to download <URL to file>`.
+    Trying to manually download the mentioned file with `curl` will also not work, and result in errors like:
+    ```
+    curl: (56) Recv failure: Connection reset by peer
+    ```
+    In all cases this was due to an intrusion prevention system scanning the associated network, and hence scanning all files going in or out of the Stratum 1.
+    Though it was a false-positive in all cases, this breaks the synchronization procedure of your Stratum 1.
+    If this is the case, you can try switching to HTTPS by using `https://aws-eu-west-s1-sync.eessi.science` for synchronizing your Stratum 1.
+    Even though there is no advantage for CVMFS itself in using HTTPS (it has built-in mechasnims for ensuring the integrity of the data),
```
casparvl, line 37: Why aren't we always recommending running with HTTPS? Is there a downside? I'd say there _might_ be a small speed penalty due to the encryption / decription. Maybe mention that this is why plain HTTP is the default.

docs/filesystem_layer/stratum1.md
```diff
@@ -1,44 +1,68 @@
...
 from the [filesystem-layer repository on GitHub](https://github.com/EESSI/filesystem-layer).
+For the commands in this section, we are assuming that you cloned this repository, and your working directory is `filesystem-layer`.
 
-Installing a Stratum 1 requires a GEO API license key, which will be used to find the (geographically) closest Stratum 1 server for your client and proxies.
-More information on how to (freely) obtain this key is available in the CVMFS documentation: https://cvmfs.readthedocs.io/en/stable/cpt-replica.html#geo-api-setup.
+!!! note GEO API
+    Installing a Stratum 1 usually requires a GEO API license key, which will be used to find the (geographically) closest Stratum 1 server for your client and proxies.
+    However, for a private Stratum 1 this can be skipped, as clients should just connect to your local Stratum 1 by default.
+    
+    If you do want to set up the GEO API, you can find more information on how to (freely) obtain this key in the CVMFS documentation: https://cvmfs.readthedocs.io/en/stable/cpt-replica.html#geo-api-setup.
```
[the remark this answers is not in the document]
casparvl: Thing is: the GEO api might not be able to tell apart e.g. if a Stratum 1 in groningen or stratum 1 in Amsterdam is closer to a client in Amsterdam, but the latency would probably show the difference. The advantage over hardcoding (as you do below) is that if the private stratum 1 is down, the fallback option is _the second closest_ (in a latency sense), which is probably the best fallback.

docs/filesystem_layer/stratum1.md
```diff
@@ -1,44 +1,68 @@
...
+    
+    You can put your license key in the local configuration file `inventory/local_site_specific_vars.yml`.
 
-You can put your license key in the local configuration file `inventory/local_site_specific_vars.yml`.
+!!! note Squid reverse proxy
+    The Stratum 1 playbooks also installs and configures a Squid reverse proxy on the server. The template configuration file for Squid can be found at `templates/eessi_stratum1_squid.conf.j2`.
```
casparvl, line 60: Wait, is that _needed_? What's the point of running a Squid next to a Stratum 1 on the same machine? The clients might as well directly connect to the Stratum 1 then, no? (probably my limited knowledge, but might be something to explain here as well)

docs/filesystem_layer/stratum1.md
```diff
@@ -1,44 +1,68 @@
...
-- set up the Stratum 1, preferably by running the Ansible playbook that we provide;
-- request a Stratum 0 firewall exception for your Stratum 1 server;
-- request a `<your site>.stratum1.cvmfs.eessi-infra.org` DNS entry;
-- open a pull request to include the URL to your Stratum 1 in the EESSI configuration.
-
-The last two steps can be skipped if you want to host a "private" Stratum 1 for your site.
-
+The EESSI project provides a number of geographically distributed public Stratum 1 servers that you can use to make EESSI available on your machine(s).
+If you want to be better protected against network outages and increase the bandwidth between your cluster nodes and the Stratum 1 servers,
+you could consider setting up a local (private) Stratum 1 server that replicates the EESSI CVMFS repository.
```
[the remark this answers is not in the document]
bedroge: That was about proxies, indeed. I could still say _multiple_ here, but the advantage of having multiple is really minimal I think.

docs/filesystem_layer/stratum1.md
```diff
@@ -47,58 +71,34 @@ sudo yum install -y ansible
...
-
-(This step is not implemented yet and can be skipped)
-
-You can request a firewall exception rule to be added for your Stratum 1 server by
-[opening an issue on the GitHub page of the filesystem layer repository](https://github.com/EESSI/filesystem-layer/issues/new).
-
-Make sure to include the IP address of your server.
-
-## Step 3: Verification of the Stratum 1
+## Verification of the Stratum 1
```
casparvl, line 99: I'd make a sub-header

### Verification of the Stratum 1 using curl

docs/filesystem_layer/stratum1.md
```diff
@@ -115,25 +115,28 @@ The second time you run it, you should get a cache hit:
...
-
 ```
 
-Example with the Norwegian Stratum 1:
+Example with the EESSI Stratum 1 running in AWS:
 
 ```bash
-curl --head http://bgo-no.stratum1.cvmfs.eessi-infra.org/cvmfs/software.eessi.io/.cvmfspublished
+curl --head http://aws-eu-central-s1.eessi.science/cvmfs/software.eessi.io/.cvmfspublished
 ```
```
casparvl, line 125: And here
```
Verification of the Stratum 1 using a client
```

That way, it is more clear that the following documentation is about _configuring that client_ (and not about configuring the stratum 1 anymore)

docs/filesystem_layer/stratum1.md
bedroge: That's indeed the main downside, as far as I know. Added a sentence about it in https://github.com/EESSI/docs/pull/157/commits/f830adb81c0ffecaaf216dec9d62bdb9f86d669e.

docs/filesystem_layer/stratum1.md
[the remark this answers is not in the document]
bedroge: I think it can only do it based on distance/geography. And since it does introduce some overhead (hopefully your local server is the geographically closest one :smile:), even the CVMFS devs recommended to switch it off in such cases.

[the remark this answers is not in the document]
bedroge: Clients should automatically connect to the local one anyway if it's the first one in the list, and if that one goes down, they will still use the GEO API of one of the public Stratum 1 servers to find the closest one.

docs/filesystem_layer/stratum1.md
bedroge: Good point. Caching of data is actually not needed, since the data is already on the same disk. I think the main usecase of this Squid is then to cache GEO API lookups, but since we recommend to disable that on private stratum 1s, this doesn't make sense. Let me check if I can easily introduce an option for this in the playbook, so that it won't set up Squid  by default, unless specifically requested (which we can then do in our playbooks for the public servers).

docs/filesystem_layer/stratum1.md
bedroge: Done in https://github.com/EESSI/docs/pull/157/commits/b018fb4a4ccdf781064658e3b9df9f1beed75d1b.

docs/filesystem_layer/stratum1.md
bedroge: Actually it does do some in-memory caching as well:
```
cache_mem 128 MB
# CERN config examples use 128 KB for both local proxies and stratum 1, but
# data objects are larger than this
maximum_object_size_in_memory 4 MB
```
That could perhaps still be somewhat beneficial...

bedroge:
> This should be more or less ready now, except that one step of our `stratum1.yml` playbook currently doesn't work due to the issue that I tried to solve with [galaxyproject/ansible-cvmfs#32](https://github.com/galaxyproject/ansible-cvmfs/pull/32).

In order to fix this (and some other issues), I've forked the repo and made the required changes: Now we just need to merge https://github.com/EESSI/filesystem-layer/pull/179 to resolve the issue.

docs/filesystem_layer/stratum1.md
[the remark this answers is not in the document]
ocaisa: You may even want to know how to "upgrade" your own S1 to an S0 so you can sync within your network

docs/filesystem_layer/stratum1.md
```diff
@@ -1,44 +1,69 @@
...
+    Trying to manually download the mentioned file with `curl` will also not work, and result in errors like:
+    ```
+    curl: (56) Recv failure: Connection reset by peer
+    ```
+    In all cases this was due to an intrusion prevention system scanning the associated network, and hence scanning all files going in or out of the Stratum 1.
+    Though it was a false-positive in all cases, this breaks the synchronization procedure of your Stratum 1.
+    If this is the case, you can try switching to HTTPS by using `https://aws-eu-west-s1-sync.eessi.science` for synchronizing your Stratum 1.
+    Even though there is no advantage for CVMFS itself in using HTTPS (it has built-in mechasnims for ensuring the integrity of the data),
+    this will prevent the described issues, as the intrusion prevention system will not be able to inspect the encrypted data.
+    As HTTPS does introduce some overhead due to the encryption/decryption, it is still recommended to use HTTP as default.
```
ocaisa, line 39: It also has a cost for us, we can't use Cloudflare caching then (but I'm not sure this belongs in our docs)

docs/filesystem_layer/stratum1.md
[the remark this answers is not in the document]
rptaylor: I don't think it is much benefit to have multiple private stratum 1 servers, but multiple proxies is a good idea.

docs/filesystem_layer/stratum1.md
```diff
@@ -1,44 +1,69 @@
...
 
-Various Linux distributions are supported, but we recommend one based on RHEL 7 or 8.
+Various Linux distributions are supported, but we recommend one based on RHEL 8 or 9.
+
+Finally, make sure that ports 80 and 8000 are open to clients.
+
+
+## Configure the Stratum 1
+
+Stratum 1 servers usually replicate from the Stratum 0 server.
```
rptaylor, line 23: "Usually"? If it did not replicate that would be a problem.

docs/filesystem_layer/stratum1.md
rptaylor: Not only is there no advantage - it should be noted this is a disadvantage because it makes caching in forward proxies impossible.

docs/filesystem_layer/stratum1.md
rptaylor: HTTPS prevents caching in forward proxies but I think it should not prevent caching in Cloudflare: https://developers.cloudflare.com/ssl/origin-configuration/ssl-modes/

docs/filesystem_layer/stratum1.md
rptaylor: There is also a typo: mechasnims

docs/filesystem_layer/stratum1.md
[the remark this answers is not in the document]
rptaylor: There have been only recent tentative ideas about implementing a latency-based mechanism, nothing concrete. Also the GeoDB can be rather fallible and inaccurate.

CVMFS_USE_GEOAPI can be enabled on a repository basis, but not by server AFAIK. So a client should do either order-based OR geo-based selection. But I think (not sure) there should be logic somewhere that a RFC1918 private network address should always be considered closest, if you are using a private IP on a local s1.

docs/filesystem_layer/stratum1.md
rptaylor: I agree with @casparvl . You could just leave the memory to the OS to help cache data for httpd. IIRC Dave recommends using a reverse proxy only for some monitoring capability that OSG uses. A priori I wouldn't expect much performance benefit - if anything it could introduce a small latency.

docs/filesystem_layer/stratum1.md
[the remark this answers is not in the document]
bedroge: https://github.com/EESSI/docs/pull/157/commits/a4851b048192da5cc7e711769a5c41607ab71e4d adds a sentence with a recommendation to at least have local proxies (bit outside the scope of this page, but probably good to mention that here as well). Left the recommendation for a stratum 1 unchanged, as I don't see much value in having more than one either.

docs/filesystem_layer/stratum1.md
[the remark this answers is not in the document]
bedroge: Reworded this part a bit in https://github.com/EESSI/docs/pull/157/commits/40668b1d66756c9513b193edff95b1b4fca88cfc.

docs/filesystem_layer/stratum1.md
bedroge: With the new version of our playbooks the Squid installation has been made optional, and it's disabled by default. So, to avoid any confusion, I've removed this part in https://github.com/EESSI/docs/pull/157/commits/7c6742ce8c88621c297a1adc565b9c72b111336d, as I don't think it should be part of this documentation.

docs/filesystem_layer/stratum1.md
bedroge: I hope you were not assuming I meant the first... :stuck_out_tongue: Anyway, reworded it a bit in https://github.com/EESSI/docs/pull/157/commits/2b32fa913c96592307fe89062740dc6f6f793ed0.

docs/getting_access/native_installation.md
```diff
@@ -69,7 +69,49 @@ The good news is that all of this only requires a handful commands :astonished:
     This is good enough for an individual client, or for testing purposes,
```
casparvl, line 69: I would:
- Remove this whole note. 
- Create a subtile right below "Native installation" with `## Installation for single clients`
- Then, at the end of the _current_ docs, make a new subtitle `## Installation for HPC (or otherwise large) systems`. Under that header, first explain that such systems are adviced to use (multiple) Squid(s) for load balancing (~1 per 500 nodes) and a full Stratum 1 for redundancy. Say that the docs here will explain how to configure the client for that, and refer to the other pages (e.g. ../filesystem_layer/stratum1.md) to learn how to setup a Stratum 1 or Squid.

bedroge: Done in https://github.com/EESSI/docs/pull/157/commits/9e97d4fc36dc185e75ef9dc0f90f7460cb536f20.

casparvl (approved):
I've followed these docs myself to setup a private stratum 1 in a VM, and a client in another VM that connected to it. This was succesfull, docs are clear to me :)

PR #158, opened 2024-03-01
Add docs on how to debug the test step that is run on the build node

casparvl:
Depends on: https://github.com/EESSI/software-layer/pull/467

https://github.com/EESSI/software-layer/pull/467 implements a test step as part of the build procedure, on the build node. Like the build, these tests can fail, and it is good to have some way to reproduce this locally.

This documentation provides instructions on how to run the test step locally. It is hard to guarantee an _exact_ match with what happens in the CI, due to the environment being a factor (e.g. number of slots that mpirun sees depends on your environment). Nevertheless, with these instructions, one should be pretty close... close enough to (probably) reproduce most test failures.

### DIFF
--- a/docs/adding_software/debugging_failed_builds.md
+++ b/docs/adding_software/debugging_failed_builds.md
@@ -190,6 +190,36 @@ After some time, this build fails while trying to build `Plumed`, and we can acc
 !!! Note
     While this might be faster than the EasyStack-based approach, this is _not_ how the bot builds. So why it _may_ reproduce the failure the bot encounters, it may not reproduce the bug _at all_ (no failure) or run into _different_ bugs. If you want to be sure, use the EasyStack-based approach.
 
+## Running the test step
+If you are still in the prefix layer (i.e. after previously building something), exit it first:
+```
+$ exit
+logout
+Leaving Gentoo Prefix with exit status 0
+```
+Then, source the EESSI init script (again):
+```
+Apptainer> source ${EESSI_CVMFS_REPO}/versions/${EESSI_VERSION}/init/bash
+Environment set up to use EESSI (2023.06), have fun!
+{EESSI 2023.06} Apptainer>
+```
+
+!!! Note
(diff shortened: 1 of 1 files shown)


docs/adding_software/debugging_failed_builds.md
```diff
@@ -190,6 +190,36 @@ After some time, this build fails while trying to build `Plumed`, and we can acc
...
+./run_tests.sh
+```
+if all goes well, you should see (part of) the EESSI test suite being run by ReFrame, finishing with something like
+
+```
+[  PASSED  ] Ran X/Y test case(s) from Z check(s) (0 failure(s), 0 skipped, 0 aborted)
+```
+
+!!! Note
+    If you are running on a system with hyperthreading enabled, you may still run into the "There are not enough slots available in the system to satisfy the X slots that were requested by the application:" error from `mpirun`, because hardware threads are not considered to be slots by default by OpenMPIs `mpirun`. In this case, run with `OMPI_MCA_hwloc_base_use_hwthreads_as_cpus=1./run_tests.sh` (for OpenMPI 4.X) or `PRTE_MCA_rmaps_default_mapping_policy=:hwtcpus ./run_tests.sh` (for OpenMPI 5.X).
```
trz42, line 221:
```suggestion
    If you are running on a system with hyperthreading enabled, you may still run into the "There are not enough slots available in the system to satisfy the X slots that were requested by the application:" error from `mpirun`, because hardware threads are not considered to be slots by default by OpenMPIs `mpirun`. In this case, run with `OMPI_MCA_hwloc_base_use_hwthreads_as_cpus=1 ./run_tests.sh` (for OpenMPI 4.X) or `PRTE_MCA_rmaps_default_mapping_policy=:hwtcpus ./run_tests.sh` (for OpenMPI 5.X).
```

trz42 (changes requested):
looks good overall

very tiny typo (missing space) ... should be easy to fix.

trz42 (approved):

PR #161, opened 2024-03-12
add available software script

laraPPr:
add script that generates the list available software and markdownpages for EESSI docs

First of three prs to add overview of software to the documentation.

PR #163, opened 2024-03-12
add the detailed software pages

laraPPr:
These pages were auto generate by the script that is added in https://github.com/EESSI/docs/pull/161


laraPPr:
I don't know how we should handle the spellcheck in this case it only found one in all the descriptions of the software. So this can easily be adjusted but in the long run the might not be manageable.

trz42 (approved):
i carefully reviewed all 279 files 😀

PR #167, opened 2024-03-22
Update partners.md

alainvanhoof:
Patrick van Brakel is no longer working at TU/e


trz42 (approved):
Thanks for the update!

PR #173, opened 2024-04-19
update outdated page on build nodes

### LINKED ISSUE #112: Software layer/Build nodes documentation out of date
The documentation at the [Build nodes](http://www.eessi.io/docs/software_layer/build_nodes/) page is out of date. Many of the installation steps have been moved to `build_container.sh` and some of the hard-coded paths have changed since the documentation was written.

Since new software installation is now achieved mainly through the bot, perhaps this page could be revised with up to date information for the purpose of manually debugging cases that require a more manual inspection of each step.

Neves-P:
Following up on an old issue and the discussion at the [support portal #58](https://gitlab.com/eessi/support/-/issues/58), this PR resolves #112.

These were the only few things that I could find that were still out of date, but perhaps there is more that I'm missing. @laraPPr could you please take a look if I missed something like we talked about on the last support meeting?

On a separate note that falls out of the scope of this PR: we might want to consider revising this page as there is some amount of overlap with the debugging failed builds [page](https://www.eessi.io/docs/adding_software/debugging_failed_builds/). Still, this page can still be useful to have a higher level overview of how builds are done on EESSI.


ocaisa (approved):
These LGTM

PR #175, opened 2024-04-30
Adding documentation on how to build and run on top of EESSI without EB

hvelab:
To follow up the short-term solution in issue https://gitlab.com/eessi/support/-/issues/48


ocaisa (changes requested):
@hvelab This is a good start. I think it is probably worth explaining the weird addition of the `GCCcore` path to `LD_LIBRARY_PATH`

docs/using_eessi/building_on_eessi.md
```diff
@@ -67,7 +67,32 @@ Finally, you should be able to load our newly build module:
 module load netCDF/4.9.0-gompi-2022a
 ```
 
-## Manually building software op top of EESSI
-Building software on top of EESSI would require your linker to use the same system-dependencies as the software in EESSI does. In other words: it requires you to link against libraries from the compatibility layer, instead of from your host OS.
+## Manually building software op top of EESSI (without EasyBuild)
+
```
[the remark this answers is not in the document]
hvelab: @ocaisa Which way do you think it's the best to add this piece of information? Should we replace the runtime part line where 
```
export LD_LIBRARY_PATH=$LIBRARY_PATH:$EBROOTGCCcore/lib64

```
is mentioned or should we add it in a separate section as an untested feature?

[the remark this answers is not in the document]
ocaisa: It's already tested in the sense that we've tried it and it works. I imagine there are probably corner cases where it will not work. I would be inclined to replace the use of `LD_LIBRARY_PATH` as this is a much more straightforward solution.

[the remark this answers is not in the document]
hvelab: just pushed a commit with this changes: [ed0c5dd](https://github.com/EESSI/docs/pull/175/commits/ed0c5dd839e380f1ba4188d7ccb6c2ea48f37b0f)

docs/using_eessi/building_on_eessi.md
```diff
@@ -67,7 +67,39 @@ Finally, you should be able to load our newly build module:
...
+The safest way to make sure all libraries will point to the required locations (and do not leak in from the host operating system) is starting an EESSI prefix shell before building. To do this: 
+
+* First of all, load the environment by starting an EESSI shell as described [here](https://www.eessi.io/docs/using_eessi/setting_up_environment). 
+* Load all dependencies you need to build your software. You must use at least a toolchain from EESSI to compile it (`foss` is a good option as it will also include MPI with OpenMPI and math libraries via FlexiBLAS/FFTW). 
+* Set manually `LD_RUN_PATH` to resolve libraries at runtime. `LIBRARY_PATH` should contain all the paths we need, and we also need to include the path to `libstdc++` from our GCC installation
+```sh
+export LD_RUN_PATH=$LIBRARY_PATH:$EBROOTGCCcore/lib64
+```
```
ocaisa, line 95: I_think_ these should be indented two spaces to line up with the bullet, GitHub doesn't let me suggest it though...

ocaisa (changes requested):
I think this looks good.

docs/using_eessi/building_on_eessi.md
hvelab: True, just fixed it in the last commit: f1fd3416f1c44834e3b15dee6e6ae6986685eff0

docs/using_eessi/building_on_eessi.md
```diff
@@ -67,7 +67,39 @@ Finally, you should be able to load our newly build module:
...
-## Manually building software op top of EESSI
-Building software on top of EESSI would require your linker to use the same system-dependencies as the software in EESSI does. In other words: it requires you to link against libraries from the compatibility layer, instead of from your host OS.
+## Manually building software op top of EESSI (without EasyBuild)
+
+!!! warning
+
+    We are working on a module file that should make building on top of EESSI (without using EasyBuild)
+    more straightforward, particularly when using `Autotools` or `CMake`. Right now, it is a little convoluted
+    and requires you to have a decent grasp of
+    * What a runtime linker (`ld-linux*.so`) is and does
```
ocaisa, line 77:
```suggestion
    * What a runtime dynamic linker (`ld-linux*.so`) is and does
```

docs/using_eessi/building_on_eessi.md
```diff
@@ -67,7 +67,39 @@ Finally, you should be able to load our newly build module:
...
+    * How to influence the behaviour of the runtime linker with `LD_LIBRARY_PATH`
+    * The difference between `LIBRARY_PATH` and `LD_LIBRARY_PATH`
+    
+    As such, this documentation is intended for "experts" in the runtime linker and it's behaviour,
+    and most cases are untested. Any feedback on this topic is highly appreciated. 
+    
+Building and running software on top of EESSI without EasyBuild is not as straightforward and requires some considerations to take care of.
```
ocaisa, line 84:
```suggestion
Building and running software on top of EESSI without EasyBuild is not straightforward and requires some considerations to take care of. 
```

docs/using_eessi/building_on_eessi.md
```diff
@@ -67,7 +67,39 @@ Finally, you should be able to load our newly build module:
...
+    * How to influence the behaviour of the runtime linker with `LD_LIBRARY_PATH`
+    * The difference between `LIBRARY_PATH` and `LD_LIBRARY_PATH`
+    
+    As such, this documentation is intended for "experts" in the runtime linker and it's behaviour,
+    and most cases are untested. Any feedback on this topic is highly appreciated. 
+    
+Building and running software on top of EESSI without EasyBuild is not as straightforward and requires some considerations to take care of. 
+
+It is expected that you will have loaded all of your required dependencies as modules from the EESSI environment. Since EESSI sets `LIBRARY_PATH` for all of the modules and the `GCC` compiler is configured to use the compat layer, there should be no additional configuration required to execute a standard build process. On the other hand, EESSI does not set `LD_LIBRARY_PATH`, so at runtime, the executable will need help finding the proper libraries that it needs. The easiest way to circumvent this is setting the environment variable `LD_RUN_PATH` during compile time as well. Thus, the dynamic linker will search for those paths when the program is executed.
```
ocaisa, line 86:
```suggestion
It is expected that you will have loaded all of your required dependencies as modules from the EESSI environment. Since EESSI sets `LIBRARY_PATH` for all of the modules and the `GCC` compiler is configured to use the compat layer, there should be no additional configuration required to execute a standard build process. On the other hand, EESSI does not set `LD_LIBRARY_PATH`, so at runtime, the executable will need help finding the proper libraries that it needs. The easiest way to circumvent this is setting the environment variable `LD_RUN_PATH` during compile time as well. Thus, the dynamic linker will search for those paths when the program is executed. 
```
```suggestion
It is expected that you will have loaded all of your required dependencies as modules from the EESSI environment. Since EESSI sets `LIBRARY_PATH` for all of the modules and the `GCC` compiler is configured to use the compat layer, there should be no additional configuration required to execute a standard build process. On the other hand, EESSI does not set `LD_LIBRARY_PATH` so, _at runtime_, the executable will need help finding the libraries that it needs to actually execute. The easiest way to circumvent this requirement is by setting the environment variable `LD_RUN_PATH` during compile time as well. With `LD_RUN_PATH` set, the program will be able to tell the dynamic linker to search in those paths when the program is being executed. 
```

docs/using_eessi/building_on_eessi.md
```diff
@@ -67,7 +67,39 @@ Finally, you should be able to load our newly build module:
...
+    As such, this documentation is intended for "experts" in the runtime linker and it's behaviour,
+    and most cases are untested. Any feedback on this topic is highly appreciated. 
+    
+Building and running software on top of EESSI without EasyBuild is not as straightforward and requires some considerations to take care of. 
+
+It is expected that you will have loaded all of your required dependencies as modules from the EESSI environment. Since EESSI sets `LIBRARY_PATH` for all of the modules and the `GCC` compiler is configured to use the compat layer, there should be no additional configuration required to execute a standard build process. On the other hand, EESSI does not set `LD_LIBRARY_PATH`, so at runtime, the executable will need help finding the proper libraries that it needs. The easiest way to circumvent this is setting the environment variable `LD_RUN_PATH` during compile time as well. Thus, the dynamic linker will search for those paths when the program is executed. 
+
+The safest way to make sure all libraries will point to the required locations (and do not leak in from the host operating system) is starting an EESSI prefix shell before building. To do this:
```
ocaisa, line 88:
```suggestion
EESSI uses a [compatibility layer](../compatibility_layer.md) to ensure that it takes as few libraries from the host as possible. The safest way to make sure all libraries will point to the required locations in the compatibility layer (and do not leak in from the host operating system) is starting an EESSI prefix shell before building. To do this: 
```

docs/using_eessi/building_on_eessi.md
```diff
@@ -67,7 +67,39 @@ Finally, you should be able to load our newly build module:
...
+    
+Building and running software on top of EESSI without EasyBuild is not as straightforward and requires some considerations to take care of. 
+
+It is expected that you will have loaded all of your required dependencies as modules from the EESSI environment. Since EESSI sets `LIBRARY_PATH` for all of the modules and the `GCC` compiler is configured to use the compat layer, there should be no additional configuration required to execute a standard build process. On the other hand, EESSI does not set `LD_LIBRARY_PATH`, so at runtime, the executable will need help finding the proper libraries that it needs. The easiest way to circumvent this is setting the environment variable `LD_RUN_PATH` during compile time as well. Thus, the dynamic linker will search for those paths when the program is executed. 
+
+The safest way to make sure all libraries will point to the required locations (and do not leak in from the host operating system) is starting an EESSI prefix shell before building. To do this: 
+
+* First of all, load the environment by starting an EESSI shell as described [here](https://www.eessi.io/docs/using_eessi/setting_up_environment).
```
ocaisa, line 90: None of these steps actually start a prefix shell! If you want to prefix shell you need to run
```
$EESSI_EPREFIX/startprefix
```
and after you do this, your environment will be completely emptied in the subshell that gets started (including EESSI), so you need to reinitialise again.

docs/using_eessi/building_on_eessi.md
```diff
@@ -67,7 +67,39 @@ Finally, you should be able to load our newly build module:
...
+
+The safest way to make sure all libraries will point to the required locations (and do not leak in from the host operating system) is starting an EESSI prefix shell before building. To do this: 
+
+* First of all, load the environment by starting an EESSI shell as described [here](https://www.eessi.io/docs/using_eessi/setting_up_environment). 
+* Load all dependencies you need to build your software. You must use at least a toolchain from EESSI to compile it (`foss` is a good option as it will also include MPI with OpenMPI and math libraries via FlexiBLAS/FFTW). 
+* Set manually `LD_RUN_PATH` to resolve libraries at runtime. `LIBRARY_PATH` should contain all the paths we need, and we also need to include the path to `libstdc++` from our GCC installation
```
ocaisa, line 92:
```suggestion
* Set manually `LD_RUN_PATH` to resolve libraries at runtime. `LIBRARY_PATH` should contain all the paths we need, and we also need to include the path to `libstdc++` from our GCC installation
```
```suggestion
* Set manually `LD_RUN_PATH` to resolve libraries at runtime. `LIBRARY_PATH` should contain all the paths we need, and we also need to include the path to `libstdc++` from our GCC installation to avoid picking up the one from the host:
```

docs/using_eessi/building_on_eessi.md
```diff
@@ -67,7 +67,39 @@ Finally, you should be able to load our newly build module:
...
+    ```sh
+        export LD_RUN_PATH=$LIBRARY_PATH:$EBROOTGCCcore/lib64
+    ```
+* Compile and make sure the library resolution points to the EESSI stack. For this, `ldd` from compatibility layer and **not** `/usr/bin/ldd` should be used when checking the binary.
```
ocaisa, line 96: Probably best to provide a running example here so people have a clearer idea of what you are talking about

docs/using_eessi/building_on_eessi.md
```diff
@@ -67,7 +67,39 @@ Finally, you should be able to load our newly build module:
...
+* Once done, exit the EESSI prefix shell (with `exit`) or use another terminal to run your software.
+
+* Run! 
+
+
+!!! Note RPATH should never point to a compatibility layer directory, only to software layer ones, as all resolving is done via the runtime linker (`ld-linux*.so`)  that is shipped with EESSI, which automatically searches these locations.
+
+The biggest downside of this approach is that your executable becomes bound to the architecture you linked your libraries for, i.e., if you add to your executable RPATH a `libhdf5.so`compiled for `intel_avx512`, you will not be able to run that binary in a machine with a different architecture. If this is an issue for you, you should look into how EESSI itself organises the location of binaries and perhaps leverage the relevant environment variables (e.g., `EESSI_SOFTWARE_SUBDIR`).
```
ocaisa, line 104:
```suggestion
The biggest downside of this approach is that your executable becomes bound to the architecture you linked your libraries for, i.e., if you add to your executable RPATH a `libhdf5.so`compiled for `intel_avx512`, you will not be able to run that binary in a machine with a different architecture. If this is an issue for you, you should look into how EESSI itself organises the location of binaries and perhaps leverage the relevant environment variables (e.g., `EESSI_SOFTWARE_SUBDIR`).
```
```suggestion
The biggest downside of this approach is that your executable becomes bound to the architecture you linked your libraries for, i.e., if you add to your executable RPATH a `libhdf5.so`compiled for `intel_avx512`, you will not be able to run that binary on a machine with a different architecture. If this is an issue for you, you should look into how EESSI itself organises the location of binaries and perhaps leverage the relevant environment variables (e.g., `EESSI_SOFTWARE_SUBDIR`).
```

ocaisa (changes requested):

docs/using_eessi/building_on_eessi.md
ocaisa: How about an MPI "Hello world" example?

docs/using_eessi/building_on_eessi.md
```diff
@@ -67,7 +67,39 @@ Finally, you should be able to load our newly build module:
...
+
+* First of all, load the environment by starting an EESSI shell as described [here](https://www.eessi.io/docs/using_eessi/setting_up_environment). 
+* Load all dependencies you need to build your software. You must use at least a toolchain from EESSI to compile it (`foss` is a good option as it will also include MPI with OpenMPI and math libraries via FlexiBLAS/FFTW). 
+* Set manually `LD_RUN_PATH` to resolve libraries at runtime. `LIBRARY_PATH` should contain all the paths we need, and we also need to include the path to `libstdc++` from our GCC installation to avoid picking up the one from the host:
+    ```sh
+        export LD_RUN_PATH=$LIBRARY_PATH:$EBROOTGCCcore/lib64
+    ```
+* Compile and make sure the library resolution points to the EESSI stack. For this, `ldd` from compatibility layer and **not** `/usr/bin/ldd` should be used when checking the binary.
+* Once done, exit the EESSI prefix shell (with `exit`) or use another terminal to run your software.
```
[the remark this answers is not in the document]
ocaisa: There are not in a prefix shell here (or at least you haven't explained to them how they get into one)

ocaisa (dismissed):
Could still be improved, but it is a good start and I don't want to hold it back any longer

ocaisa (approved):

PR #181, opened 2024-06-04
Update scorecards version

ocaisa:
Update to 2.3.3, based on matching to https://github.com/ossf/scorecard/blob/main/.github/workflows/scorecard-analysis.yml


ocaisa:
This would stop running scorecard on PRs, it's what they currently do upstream

ocaisa:
This is working, although the upload seems to be a problem: https://github.com/EESSI/docs/actions/runs/9363017577

boegel (approved):

PR #184, opened 2024-06-04
improve error reporting when available_software.py script is run in incorrect environment

boegel:
With these changes, the script produces better error messages when it's being run when:

* EESSI is not available (`/cvmfs/software.eessi.io/...` does not exist);
* Lmod is not properly set up (`$LMOD_CMD` does not specify path to Lmod binary);

When `$MODULEPATH` is empty, the script also won't crash anymore now.


laraPPr (approved):
lgtm

PR #187, opened 2024-06-04
Fix CI for available_software.py in script_module_list.yml 

laraPPr:


laraPPr:
This is still not triggering the tests. And I think it should since a file is changed in the paths.

boegel:
Changes probably only take effect when they're active in `main` branch, so merging...

PR #188, opened 2024-06-04
Host injections, lmod hooks and moving gpu-related host-injection instructions

casparvl:
Currently, the `host_injections` directory was introduced as part of the `Advanced Usage => Accelerator support => GPUs`. I think we better introduce the concept of `host_injections` being a location that can be used for site-specific EESSI configuration first. Since this is about _configuration_, it makes sense to me to do this as part of the `Installation` instructions. I also imagine that the `Installation` chapter would be the only chapter that sysadmins read, so in my opinion it should contain everything needed to install _and_ configure EESSI to it's full functionality (including GPU support, MPI tuning, etc).

I've also added new documentation on how to write `Lmod` hooks in `host_injections`. This basically provides the documentation for the functionality added in [this PR](https://github.com/EESSI/software-layer/pull/525).


mkdocs.yml
```diff
@@ -26,6 +26,10 @@ nav:
     - Is EESSI already installed?: getting_access/is_eessi_accessible.md
     - Native: getting_access/native_installation.md
     - Container: getting_access/eessi_container.md
+    - Configuring EESSI:
```
xinan1911, line 29: How about calling it _Advanced configuration_?

xinan1911: And adjust the "Installation" to "Installation and advance configuration"

docs/site_specific_config/host_injections.md
```diff
@@ -0,0 +1,43 @@
+# How to configure EESSI
+
+## The `host_injections` variant symlink
+
+While the EESSI software stack aims to work 'out of the box', one cannot avoid some degree of site-specific configuration for certain functionality and/or performance tuning. For example, GPU drivers may be in different locations on different systems. Also, a site might want to tune the OpenMPI installation that comes with EESSI for optimal performance on the local hardware.
```
xinan1911, line 5: The first sentence seems to be counter-intuitive to what EESSI tries to achieve. It is more reasonable to say that due to the copyrights of the software or freedom of configuration, the site specific configuration with host injection is enabled in EESSI.

casparvl: > The first sentence seems to be counter-intuitive to what EESSI tries to achieve

Not really. The primary goal is to offer the same, optimized software stack everywhere. Sure, it is a benefit if it's easy to deploy EESSI, as that will increase uptake. And where possible, we abstract from host-specific differences. But there _are_ differences between hosts that we simply have to deal with through configuration, the prime example being the location of the GPU drivers for GPU functionality.

I'll think about the phrasing though, maybe I shouldn't start with the statement that we want things to work out of the box - that might simply be too optimistic.

mkdocs.yml
casparvl: I agree on including the configuration part in the top header, but I wouldn't call it `Advanced`. Sure, you could argue that site-specific LMOD hooks to tune MPI installations are a bit advanced. But e.g. symlinking the GPU drivers is a) super easy and b) very necessary on any GPU capable system. I don't want to imply that readers can just 'skip' this configuration stuff unless they need something _very_ special (and that's what the word 'advanced' seems to imply - that it's not needed for 'regular'  useage. But it is when it comes to GPUs).

docs/site_specific_config/host_injections.md
casparvl: Ok, made an additional (very short) section explaining when/why configuration i needed. Also, removed the phrasing which states that we aim to work 'out-of-the-box', that's simply not feasible, so no point in mentioning that.

xinan1911 (approved):
Looks good to me!

PR #189, opened 2024-06-05
add workflow to update overview of available software

boegel:
For now, I've only listed `workflow_dispatch` as a trigger, which essentially means that triggering the creation of a PR will only happen when someone runs the workflow via https://github.com/EESSI/docs/blob/main/.github/workflows/update_available_software.yml .

I prefer taking that approach initially, before we enable daily (or whatever) runs of this workflow to automatically get a PR opened when the overview of available software needs to be updated...

The token (available as GitHub Action environment secret in this repo as `EESSIBOT_GITHUB_TOKEN`) is currently valid until 4 Aug'24, we'll need to make sure it gets replaced regularly (out of abundance of care).

Example of a pull request that is produced by this workflow (which was tweaked to use the `boegelbot` account and target my fork @ `boegel/EESSI-docs`: https://github.com/boegel/EESSI-docs/pull/2

### DIFF
--- a/scripts/available_software/available_software.py
+++ b/scripts/available_software/available_software.py
@@ -17,7 +17,6 @@ import re
 import subprocess
 import sys
 import time
-from pathlib import Path
 from typing import Union, Tuple
 import numpy as np
 from mdutils.mdutils import MdUtils
@@ -35,11 +34,7 @@ EXCLUDE_CPU_TARGETS = ['x86_64/amd/zen4']
 
 def main():
     os.environ["SHELL"] = "/bin/bash"
-    current_dir = Path(__file__).resolve()
-    project_name = 'docs'
-    root_dir = next(
-        p for p in current_dir.parents if p.parts[-1] == project_name
-    )
+    root_dir = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
     path_data_dir = os.path.join(root_dir, "docs/available_software/data")
 
     # Generate the JSON overviews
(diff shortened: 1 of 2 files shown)


scripts/available_software/available_software.py
```diff
@@ -35,11 +34,7 @@
 
 def main():
     os.environ["SHELL"] = "/bin/bash"
-    current_dir = Path(__file__).resolve()
-    project_name = 'docs'
-    root_dir = next(
-        p for p in current_dir.parents if p.parts[-1] == project_name
-    )
+    root_dir = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
boegel, line 37: I changed this because the `available_software.py` script was failing in a checkout [my fork](https://github.com/boegel/EESSI-docs), because the top-level directory there was `EESSI-docs` rather than `docs`.

I don't see the need to use `pathlib` here, it makes the code a whole lot more messy and hard to understand, imho

trz42 (changes requested):
Nice! I'd suggest to use specific commits for the actions (and allow only those if they are not yet allowed), but other than that no specific comments.

boegel:
> Nice! I'd suggest to use specific commits for the actions (and allow only those if they are not yet allowed), but other than that no specific comments.

Good point, fixed in 9ce8952

trz42 (approved):
Let's try it 🫣

ocaisa:
Almost
> The overview below shows which WRF installations are available per HPC-UGent Tier-2cluster,

ocaisa:
Ah dispatch workflow, but still problems: https://github.com/EESSI/docs/actions/runs/9391493091/job/25863767460

PR #190, opened 2024-06-05
Incorrect hashes for update available software action

ocaisa:


trz42 (changes requested):
Tiny version typo...

trz42 (approved):
Thanks for the fixes!

PR #195, opened 2024-06-06
add script to check whether JSON data for available software has been updated (ignoring timestamp)

boegel:

PR #196, opened 2024-06-06
Use hash for flake8 action

ocaisa:


ocaisa:
Why didn't this run and fail for https://github.com/EESSI/docs/pull/195 ?

ocaisa:
@laraPPr Do you have any idea, why would it run on push to main but not on the PR?

boegel:
@ocaisa Maybe path should be `./scripts/**` ? Doesn't make sense why it did trigger for `main` though...

I'll merge this since it should at least fix the problem we see with the CI workflows on `main`

boegel (approved):

PR #197, opened 2024-06-06
Correct installation of requirements for updating available software

ocaisa:


.github/workflows/update_available_software.yml
```diff
@@ -24,13 +24,13 @@ jobs:
...
           . venv/bin/activate
-          pip install -r requirements.txt
+          pip install -r scripts/available_software/requirements.txt
 
           # copy current JSON data, so we can compare after updating it
           cp docs/available_software/data/json_data.json docs/available_software/data/json_data.json.orig
           cp docs/available_software/data/json_data_detail.json docs/available_software/data/json_data_detail.json.orig
 
-          python scripts/available_software/available_software.py
+          python scripts/available_software/requirements.txt
```
trz42, line 33: Is this change intended?

trz42 (changes requested):
Running
```
python scripts/available_software/requirements.txt
```
is maybe not intended?

.github/workflows/update_available_software.yml
ocaisa: Nope! Whoops, wrong spot

ocaisa:
Hmm, closing this, but my impression is it is installing the wrong `requirements.txt`, but I can't see where

ocaisa:
Ah, I had one thing right and one wrong!

.github/workflows/update_available_software.yml
ocaisa: Fixed!

trz42 (approved):

boegel:
Ah, overlooked this in #195, thanks!

PR #199, opened 2024-06-06

EESSIbot:


boegel:
@ocaisa Let's try to trigger the workflow again now, shouldn't open a new PR since all available software is already included in the docs (cfr. #195)

ocaisa:
@boegel Seems to work as designed: https://github.com/EESSI/docs/actions/runs/9398000717/job/25882416283

PR #215, opened 2024-06-25

EESSIbot:


boegel:
There's no point in already including `aarch64/a64fx` in the overview already, we should exclude it for now, see also https://github.com/EESSI/docs/pull/216

PR #69, opened 2021-04-06
WIP: documentation for installing+configuring CVMFS clients for EESSI

bedroge:

PR #234, opened 2024-07-30
add note about cache directory

trz42:
This PR adds a small note to the description of the CernVM-FS client installation. It is taken from #69 which could be closed after this PR here got merged.


ocaisa (approved):
LGTM

PR #245, opened 2024-08-02
Add information on how to use EESSI in CI

ocaisa:


ocaisa:
Build of this available at https://ocaisa.github.io/docs/using_eessi/eessi_in_ci/

docs/using_eessi/eessi_in_ci.md
```diff
@@ -0,0 +1,49 @@
...
+EESSI is already available as both a GitHub Action and a GitLab CI/CD component, which means you can easily integrate
+it if you use continuous integration within those ecosystems.
+
+!!! note
+
+    Both of these EESSI CI tools support the use of [`direnv`](https://direnv.net/) to allow you to store your desired
+    environment within a `.envrc` file within your repository. See the documentation of the individual tools for
+    detailed usage.
+
+## The EESSI GitHub Action
```
boegel, line 12: Would it be overkill to have separate pages for GitHub Actions & GitLab, if only so we can have a nice menu item for both?

ocaisa: I can add menu items using the heading anchors if that's desired. I would prefer not to split the page as ultimately I would like the CI to behave the same on all platforms and we can have a definitive place to outline that behaviour.

A good example of why to do it that way is the behaviour of `direnv`. `direnv` works great, you can have any kind of scripting you want in the `.envrc` and (like how Lmod does it) it will run the script and give you the set the environment variables from that environment. However, it _only_ sets environment variables (so no functions), which means if you use it without Lmod initiated then you still can't use Lmod when it is done. I sidestep this right now by initiating the EESSI environment before  using `direnv` and this makes life even easier as `direnv` is already in the our compat layer. Having to explain that multiple times (and keep them in sync) would be a pain.

boegel:
I think @Neves-P volunteerd to review this in detail?

Neves-P:
> I think @Neves-P volunteerd to review this in detail?

Yes, indeed 👍

docs/using_eessi/eessi_in_ci.md
```diff
@@ -0,0 +1,49 @@
...
+    - name: Test EESSI
+      run: |
+        module avail
+      shell: bash
+```
+
+## The EESSI GitLab CI/CD component
+
+The EESSI GitLab CI/CD component can be found in the [GitLab CI/CD Catalog](https://gitlab.com/explore/catalog), at
+[https://gitlab.com/explore/catalog/eessi/gitlab-eessi](https://gitlab.com/explore/catalog/eessi/gitlab-eessi) .
```
Neves-P, line 36:
```suggestion
[https://gitlab.com/explore/catalog/eessi/gitlab-eessi](https://gitlab.com/explore/catalog/eessi/gitlab-eessi) .
[https://gitlab.com/explore/catalog/eessi/gitlab-eessi](https://gitlab.com/explore/catalog/eessi/gitlab-eessi).
```

docs/using_eessi/eessi_in_ci.md
```diff
@@ -0,0 +1,49 @@
...
+!!! note
+
+    Both of these EESSI CI tools support the use of [`direnv`](https://direnv.net/) to allow you to store your desired
+    environment within a `.envrc` file within your repository. See the documentation of the individual tools for
+    detailed usage.
+
+## The EESSI GitHub Action
+
+The EESSI GitHub Action can be found on the [GitHub Marketplace](https://github.com/marketplace),
+at [https://github.com/marketplace/actions/eessi](https://github.com/marketplace/actions/eessi) .
```
Neves-P, line 15:
```suggestion
at [https://github.com/marketplace/actions/eessi](https://github.com/marketplace/actions/eessi) .
at [https://github.com/marketplace/actions/eessi](https://github.com/marketplace/actions/eessi).
```

Neves-P (changes requested):
I think the contents are clear and have all the necessary information! 

One thing I'd suggest adding is an entry for "EESSI in CI" or something like that under "Basic usage" on the menu bar on the left so that this functionality is easier to find. Otherwise, my only suggestions are removing a couple of extra spaces.

Neves-P (approved):
Looks good to me!

PR #256, opened 2024-08-07
Use a time template in module documentation generation

### LINKED ISSUE #255: Centralise the date for the software module descriptions
Right now our PRs for updating the overview of available software (like https://github.com/EESSI/docs/pull/254) are unreviewable because every file gets an updated string
```
*(This data was automatically generated on Wed, 07 Aug 2024 at 12:36:59 UTC)* 
```
Rather than do it this way, can we have the date as part of our MkDocs configuration file and just use a variable in those files. This can be done for example using https://github.com/fralau/mkdocs-macros-plugin.

That approach would mean meaningful PRs (and a smaller git repo!).

ocaisa:
Fixes #255

### DIFF
--- a/.github/workflows/update_available_software.yml
+++ b/.github/workflows/update_available_software.yml
@@ -1,6 +1,5 @@
 name: Update overview of available software in EESSI
 on:
-  pull_request:
   workflow_dispatch:
   schedule:
     # run every 4 hours
@@ -40,7 +39,6 @@ jobs:
           ./scripts/update_generated_time.sh mkdocs.yml
           
           git status
-          git diff docs/available_software/detail/zstd.md
 
           # determine whether pull request should be opened:
           # if JSON files in docs/available_software/data have been updated, then a PR should be opened
--- a/.github/workflows/update_available_software.yml
+++ b/.github/workflows/update_available_software.yml
@@ -37,6 +37,8 @@ jobs:
 
           export TIME_GENERATED_TEMPLATE="{{ generated_time }}"
           python scripts/available_software/available_software.py
+          ./scripts/update_generated_time.sh mkdocs.yml
+          
           git status
 
           # determine whether pull request should be opened:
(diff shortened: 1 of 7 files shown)


.github/workflows/update_available_software.yml
```diff
@@ -1,5 +1,6 @@
 name: Update overview of available software in EESSI
 on:
+  pull_request:
```
ocaisa, line 3: This is temporary so I can test this

ocaisa:
Ok, this is ready for showtime

mkdocs.yml
```diff
@@ -123,6 +125,7 @@ extra:
   social:
     - icon: fontawesome/brands/twitter
       link: https://twitter.com/eessi_hpc
+  generated_time: "Wed, 07 Aug 2024 at 17:08:54 CEST"
```
boegel, line 128: I would add a comment here that this gets auto-updated via the script:
```suggestion
  # this gets auto-updated via update_generated_time.sh script run in update_available_software.yml action
  generated_time: "Wed, 07 Aug 2024 at 17:08:54 CEST"

File: .github/workflows/update_available_software.yml [hunk 3]
```diff
@@ -34,7 +34,10 @@ jobs:
           cp docs/available_software/data/json_data.json docs/available_software/data/json_data.json.orig
           cp docs/available_software/data/json_data_detail.json docs/available_software/data/json_data_detail.json.orig
 
+          export TIME_GENERATED_TEMPLATE="{{ generated_time }}"
```
boegel, line 37: Why the side-step here via this environment variable, why not just "hardcode" `{{ generated_time }}` in the generated pages?

boegel (changes requested):

.github/workflows/update_available_software.yml
ocaisa: So that others sites can use the script without being forced to take the same approach.

boegel: OK, doesn't hurt I guess.

boegel (approved):

PR #259, opened 2024-08-09

EESSIbot:


ocaisa (approved):

ocaisa:
This includes the update to `mkdocs.yml` so should be good to go (and the last time we see these monster PRs!)

ocaisa:
Working as expected, see shift to reporting in UTC (from CEST) in https://www.eessi.io/docs/available_software/detail/Abseil/ for example

PR #262, opened 2024-08-13
Add section on writing portable tests for the EESSI test suite

casparvl:
Work in progress. Ideally, I need API documentation first. Or, if this proves difficult, we should probably make a note that this will come soon.

### DIFF
--- a/docs/test-suite/writing-portable-tests.md
+++ b/docs/test-suite/writing-portable-tests.md
@@ -2,7 +2,7 @@
 
 This page is a tutorial on how to write a new test for the [EESSI test suite](https://github.com/EESSI/test-suite).
 
-If you already know how to write regular ReFrame tests, we suggest you read the [High-level overview](#high-level-overview) and [Test requirements](#test-requirements) sections, then skip ahead to [developing a portable ReFrame test](#developing-a-portable-reframe-test).
+If you already know how to write regular ReFrame tests, we suggest you read the [High-level overview](#high-level-overview) and [Test requirements](#test-requirements) sections, then skip ahead to [Step 3: implementing as a portable ReFrame test](#as-portable-reframe-test).
 
 ## High-level overview
 
@@ -68,8 +68,8 @@ mpirun -np 256 python3 mpi4py_reduce.py --n_iter 1000 --n_warmup 100
 
 Clearly, such job scripts are not very portable: these only work on SLURM systems, we had to duplicate a lot to run on different scales, we would have to duplicate even more if we wanted to run test `mpi4py` versions, etc. This is where `ReFrame` comes in: it has support for different schedulers, and allows one to easily specify a range of parameters (such as the number of tasks in the above example) to create tests for.
 
-### Step 2: implementing it as a non-portable test in ReFrame
-First, let us implement this as a non-portable test in ReFrame. We will not elaborate on how to write ReFrame tests, it is well-documented in the official [ReFrame documentation](https://reframe-hpc.readthedocs.io/en/stable/index.html). We have put extensive comments in the test definition below, to make it easier to understand when you have limited familiarity with ReFrame. Whenever the variables below have a specific meaning in ReFrame, we referenced the official documentation:
+### Step 2: implementing as a non-portable ReFrame test
+First, let us implement this as a non-portable test in ReFrame. This code can be found under `tutorials/mpi4py/mpi4py_system_specific.py` in the [EESSI test suite](https://github.com/EESSI/test-suite/) repository. We will not elaborate on how to write ReFrame tests, it is well-documented in the official [ReFrame documentation](https://reframe-hpc.readthedocs.io/en/stable/index.html). We have put extensive comments in the test definition below, to make it easier to understand when you have limited familiarity with ReFrame. Whenever the variables below have a specific meaning in ReFrame, we referenced the official documentation:
 
 ```python
 """
@@ -170,11 +170,11 @@ This single test class will generate 6 test instances: tests with 2, 128 and 256
 
 This test _works_, but is _not_ very portable. If we move to a system with 192 cores per node, the current `scale` parameter is a bit awkward. The test would still run, but we wouldn't have a test instance that just tests this on a full (single) node or a full two nodes. Furthermore, if we add a new `mpi4py` module in EESSI, we would have to alter the test to add the name to the list, since the module names are hardcoded in this test.
 
-### Developing a portable ReFrame test
+### Step 3: implementing as a portable ReFrame test {#as-portable-reframe-test}
 
 In the previous section, there were several system-specific items in the test. In this section, we will show how we use the EESSI hooks to avoid hard-coding system specific information. We do this by looking at the test step-by-step.
 
-#### Replacing hard-coded test scales
+#### Replacing hard-coded test scales (mandatory)
 
 We replace the hard-coded
 
@@ -241,7 +241,7 @@ The first hook ([set_tag_scale](TODO: API reference)) sets a number of custom at
 
 The second hook, `assign_tasks_per_compute_unit`, is used to set the task count. This hook sets the `self.num_tasks` and `self.num_tasks_per_node` we hardcoded before. In addition, it sets the `self.num_cpus_per_task`. In this case, we call it with the `COMPUTE_UNIT[CPU]` argument, which means one task will be launched per (physical) CPU available. Thus, for the `1_node` scale, this would run the `mpi4py` test with 128 tasks on a 128-core node, and with 192 tasks on a 192-core node. Check the [API reference](TODO) for other valid `COMPUTE_UNIT`'s.
 
-#### Replacing hard-coded module names
+#### Replacing hard-coded module names (mandatory)
 
 If we write an `mpi4py` test, we typically want to run this for _all_ `mpi4py` modules that are currently on our `$MODULEPATH`. We do that by replacing
 
@@ -275,7 +275,7 @@ by
 
 This hook assumes that `self.module_name` has been set, but has the added advantage that a user running the EESSI test suite can overwrite the modules to load from the command line when running ReFrame (see [Overriding test parameters](https://www.eessi.io/docs/test-suite/usage/#overriding-test-parameters-advanced)).
 
-#### Replacing hard-coded valid_systems
+#### Replacing hard-coded valid_systems (mandatory)
 
 The `valid_systems` attribute is a mandatory attribute to specify in a ReFrame test. However, we can set it to match any system:
 
@@ -308,13 +308,13 @@ from eessi.testsuite.constants import FEATURES, GPU, GPU_VENDOR, GPU_VENDORS, NV
 },
 ```
 
-In practice, one will rarely hard-code this `valid_systems` string. Instead, we have a hook [filter_valid_systems_by_device_type](TODO API REF). It does the above, and a bit more: it also checks if the module that the test is generated for is CUDA-enabled (in case of a test for `NVIDIA` GPUs), and _only then_ will it generate a GPU-based test. This hook is mandatory for all tests (even if just to declare they need a CPU to run).
+In practice, one will rarely hard-code this `valid_systems` string. Instead, we have a hook [filter_valid_systems_by_device_type](TODO API REF). It does the above, and a bit more: it also checks if the module that the test is generated for is CUDA-enabled (in case of a test for `NVIDIA` GPUs), and _only then_ will it generate a GPU-based test. Calling this hook is mandatory for all tests (even if just to declare they need a CPU to run).
 
-Another aspect is that not all systems may be able to run tests of all of the standard `SCALES`. Thus, a test can also declare it needs a certain _scale_. I.e. a test for the `16_nodes` scale clearly needs a partition with at least 16 nodes. This is taken care of by the [filter_supported_scales](TODO API REF) hook. This hook is mandatory for all tests.
+Another aspect is that not all systems may be able to run tests of all of the standard `SCALES`. Thus, a test can also declare it needs a certain _scale_. I.e. a test for the `16_nodes` scale clearly needs a partition with at least 16 nodes. This is taken care of by the [filter_supported_scales](TODO API REF) hook. Calling this hook is also mandatory for all tests.
 
 There may be other hooks that facilitate valid system selection for your tests, but please check the [API documentation](TODO: INSERT REFERENCE TO API DOCS) for a full list.
 
-#### Requesting sufficient memory
+#### Requesting sufficient memory (mandatory)
 
 When developing the test, we don't know how much memory the node will have on which it will run. However, we _do_ know how much our application _needs_.
 
@@ -365,6 +365,8 @@ P: max_mem_in_mib: 403 MiB (r:0, l:None, u:None)
 P: max_mem_in_mib: 195 MiB (r:0, l:None, u:None)
 ```
 
+If you are _not_ on a system where your scheduler runs jobs in cgroups, you will have to figure out the memory consumption in another way (e.g. by checking memory useage in `top` while running the test).
+
 We now have a pretty good idea how the memory per node scales: for our smallest process counts (1 core), it's about 200 MiB per process, while for our largest process counts (16 nodes, 16*192 processes), it's 22018 MiB per node (or about 115 MiB per process). If we wanted to do really well, we could define a linear function (with offset) and fit it through the data (and round up to be on the safe side, i.e. make sure there is _enough_ memory). Then, we could call the hook like this:
 
 ```python
@@ -384,7 +386,7 @@ def request_mem(self):
 ```
 Note that requesting too high an amount of memory means the test will be skipped on nodes that cannot meet that requirement (even if they might have been able to run it without _actually_ running out of memory). Requesting too little will risk nodes running out of memory while running the test. Note that many HPC systems have an amount memory of around 1-2 GB/core. It's good to ensure (if you can) that the memory requests for all valid `SCALES` for your test do not exceed the total amount of memory available on typical nodes.
 
-#### Requesting task and/or process binding
+#### Requesting task and/or process binding (optional)
 
 Binding processes to a set of cores prevents the OS from migrating such processes to other cores. Especially on multi-socket systems, process migration can cause performance hits, especially if a process is moved to a CPU core on the other socket. Since this is controlled by the OS, and dependent on what other processes are running on the node, it may cause unpredictable performance: in some runs, processes might be migrated, while in others, they aren't.
 
@@ -410,11 +412,11 @@ def set_binding(self):
 
 The use of this hook is optional. Note that thread binding can sometimes cause unwanted behaviour: even if e.g. 8 cores are allocated to the process and 8 threads are launched, we have seen codes that bind all those threads to a single core (e.g. core 0) when core binding is enabled. Please verify that enabling core binding does not introduce any unwanted binding behaviour for your code.
 
-#### Defining OMP_NUM_THREADS
+#### Defining OMP_NUM_THREADS (optional)
 
 The `set_omp_num_threads` hook sets the `OMP_NUM_THREADS` environment variable based on the number of `cpus_per_task` defined in the ReFrame test (which in turn is typically set by the `assign_tasks_per_compute_unit` hook). For OpenMP codes, it is generally recommended to call this hook, to ensure they launch the correct amount of threads.
 
-### skip_if's
+### Skipping tests instances when required (optional)
 Preferably, we prevent test instances from being generated (i.e. before ReFrame's `setup` phase) if we know that they cannot run on a certain system. However, sometimes we need information on the nodes that will run it, which is only available _after_ the `setup` phase. That is the case for anything where we need information from e.g. the [reframe.core.pipeline.RegressionTest.current_partition](https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html#reframe.core.pipeline.RegressionTest.current_partition). The `assign_tasks_per_compute_unit` hook for example needs uses this property to get the core count of a node, and thus needs to be executed after the `setup` phase.
 
 For example, we might know that a test case only scales to around 300 tasks, and above that, execution time increases rapidly. In that case, we'd want to skip any test instance that results in a larger amount of tasks, but we only know this after `assign_tasks_per_compute_unit` has been called. E.g. the `2_nodes` scale would run fine on systems with 128 cores per node, but would exceed the task limit of 300 on systems with `192` cores per node.
@@ -464,7 +466,7 @@ An example of such a collective operation is the [MPI_REDUCE](https://www.mpi-fo
 #### The mpi4py test
 In this example, we will implement a test that does an `MPI_Reduce` on the rank, using the sum operation. This makes it easy to validate the result, as we know that for N processes, the theoretical sum of all ranks (0, 1, ... N-1) is `(N * (N-1) / 2)`.
 
-Our initial code is a python script `mpi4py_reduce.py`:
+Our initial code is a python script `mpi4py_reduce.py`, which can be found in `tutorials/mpi4py/src/mpi4py_reduce.py` in the [EESSI test suite](https://github.com/EESSI/test-suite/) repository:
 ```python
 #!/usr/bin/env python
 """
--- a/docs/test-suite/writing-portable-tests.md
+++ b/docs/test-suite/writing-portable-tests.md
@@ -100,6 +100,9 @@ class EESSI_MPI4PY(rfm.RunOnlyRegressionTest):
     valid_systems = ['snellius']
 
     # ReFrame will generate a test for each module
+    # NOTE: each parameter adds a new dimension to the parametrization space. 
+    # (EG 4 parameters with (3,3,2,2) possible values will result in 36 tests).
+    # Be mindful of how many parameters you add to avoid the number of tests generated being excessive.
     # https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html#reframe.core.builtins.parameter
     module_name = parameter(['mpi4py/3.1.4-gompi-2023a', 'mpi4py/3.1.5-gompi-2023b'])
 
--- a/docs/test-suite/writing-portable-tests.md
+++ b/docs/test-suite/writing-portable-tests.md
@@ -170,7 +170,7 @@ This single test class will generate 6 test instances: tests with 2, 128 and 256
 
 This test _works_, but is _not_ very portable. If we move to a system with 192 cores per node, the current `scale` parameter is a bit awkward. The test would still run, but we wouldn't have a test instance that just tests this on a full (single) node or a full two nodes. Furthermore, if we add a new `mpi4py` module in EESSI, we would have to alter the test to add the name to the list, since the module names are hardcoded in this test.
 
-### Step 3: implementing as a portable ReFrame test {#as-portable-reframe-test}
+### Step 3: implementing as a portable ReFrame test { #as-portable-reframe-test }
 
 In the previous section, there were several system-specific items in the test. In this section, we will show how we use the EESSI hooks to avoid hard-coding system specific information. We do this by replacing the system-specific parts of the test from Step 2 bit by bit. The full final test can be found under `tutorials/mpi4py/mpi4py_portable.py` in the [EESSI test suite](https://github.com/EESSI/test-suite/) repository.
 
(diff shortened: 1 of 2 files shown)


casparvl:
As a note: seeing the mandatory hooks, I think we could opt for a [RegressionMixin class](https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html#reframe.core.pipeline.RegressionMixin) to make things simpler, and simply make it mandatory to inherit from that. However, that's more a long-term effort, I don't want to block people from contributing until we sort this out - it might be months, it might be never.

docs/test-suite/writing-portable-tests.md
```diff
@@ -0,0 +1,526 @@
...
+
+#### Defining OMP_NUM_THREADS (optional)
+
+The `set_omp_num_threads` hook sets the `OMP_NUM_THREADS` environment variable based on the number of `cpus_per_task` defined in the ReFrame test (which in turn is typically set by the `assign_tasks_per_compute_unit` hook). For OpenMP codes, it is generally recommended to call this hook, to ensure they launch the correct amount of threads.
+
+#### Skipping tests instances when required (optional)
+Preferably, we prevent test instances from being generated (i.e. before ReFrame's `setup` phase) if we know that they cannot run on a certain system. However, sometimes we need information on the nodes that will run it, which is only available _after_ the `setup` phase. That is the case for anything where we need information from e.g. the [reframe.core.pipeline.RegressionTest.current_partition](https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html#reframe.core.pipeline.RegressionTest.current_partition). The `assign_tasks_per_compute_unit` hook for example needs uses this property to get the core count of a node, and thus needs to be executed after the `setup` phase.
```
smoors, line 422:
```suggestion
Preferably, we prevent test instances from being generated (i.e. before ReFrame's `setup` phase) if we know that they cannot run on a certain system. However, sometimes we need information on the nodes that will run it, which is only available _after_ the `setup` phase. That is the case for anything where we need information from e.g. the [reframe.core.pipeline.RegressionTest.current_partition](https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html#reframe.core.pipeline.RegressionTest.current_partition). The `assign_tasks_per_compute_unit` hook for example uses this property to get the core count of a node, and thus needs to be executed after the `setup` phase.
```

docs/test-suite/writing-portable-tests.md
```diff
@@ -0,0 +1,526 @@
...
+
+For example, we might know that a test case only scales to around 300 tasks, and above that, execution time increases rapidly. In that case, we'd want to skip any test instance that results in a larger amount of tasks, but we only know this after `assign_tasks_per_compute_unit` has been called. E.g. the `2_nodes` scale would run fine on systems with 128 cores per node, but would exceed the task limit of 300 on systems with `192` cores per node.
```
smoors, line 424:
```suggestion
For example, we might know that a test only scales to around 300 tasks, and above that, execution time increases rapidly. In that case, we'd want to skip any test instance that results in a larger amount of tasks, but we only know this after `assign_tasks_per_compute_unit` has been called. E.g. the `2_nodes` scale would run fine on systems with 128 cores per node, but would exceed the task limit of 300 on systems with `192` cores per node.
```

docs/test-suite/writing-portable-tests.md
```diff
@@ -0,0 +1,526 @@
...
+
+We can skip any generated test instances using the `skip_if` function. E.g. to skip the test if the total task count exceeds 300, we'd need to call `skip_if` _after_ the task count has been set by `assign_tasks_per_compute_unit`:
```
smoors, line 426:
```suggestion
We can skip any generated test cases using the `skip_if` function. E.g. to skip the test if the total task count exceeds 300, we'd need to call `skip_if` _after_ the task count has been set by `assign_tasks_per_compute_unit`:
```

docs/test-suite/writing-portable-tests.md
```diff
@@ -0,0 +1,526 @@
...
+
+# Warmup
+t0 = time.time()
+for i in range(n_warmup):
+    total = MPI.COMM_WORLD.reduce(rank)
+
+# Actual reduction, multiple iterations for accuracy of timing
+t1 = time.time()
+for i in range(n_iter):
+    total = MPI.COMM_WORLD.reduce(rank)
```
smoors, line 508: see https://github.com/EESSI/test-suite/pull/149/files#r1721635470

docs/test-suite/writing-portable-tests.md
```diff
@@ -0,0 +1,526 @@
...
+To understand what this test does, you need to know some basics of MPI. If you know about MPI, you can skip this section.
+
+The MPI standard defines how to communicate between multiple processes that work on a common computational task. Each process that is part of the computational task gets a unique identifier (0 to N-1 for N processes), the MPI rank, which can e.g. be used to distribute a workload. The MPI standard defines communication between two given processes (so-called point-to-point communication), but also between a set of N processes (so-called collective communication).
+
+An example of such a collective operation is the [MPI_REDUCE](https://www.mpi-forum.org/docs/mpi-4.1/mpi41-report/node130.htm#Node130) call. It reduces data elements from multiple processes with a certain operation, e.g. it takes the sum of all elements or multiplication of all elements.
+
+#### The mpi4py test
+In this example, we will implement a test that does an `MPI_Reduce` on the rank, using the sum operation. This makes it easy to validate the result, as we know that for N processes, the theoretical sum of all ranks (0, 1, ... N-1) is `(N * (N-1) / 2)`.
```
smoors, line 469:
```suggestion
In this example, we will implement a test that does an `MPI_Reduce` on the rank, using the `MPI.SUM` operation. This makes it easy to validate the result, as we know that for N processes, the theoretical sum of all ranks (0, 1, ... N-1) is `(N * (N-1) / 2)`.
```

docs/test-suite/writing-portable-tests.md
```diff
@@ -0,0 +1,526 @@
...
+
+Assuming we have `mpi4py` available, we could run this manually using
+```
+$ mpirun -np 4 python3 mpi4py_reduce.py
+Total ranks: 4
+Sum of all ranks: 6
+Time elapsed: 3.609e-06
+```
+
+This started 4 processes, with ranks 0, 1, 2, 3, and then summed all the ranks (`0+1+2+3=6`) on the process with rank 0, which finally printed all this output. The whole reduction operation is performed `n_iter` times, so that we get a more accurate timing.
```
smoors, line 526:
```suggestion
This started 4 processes, with ranks 0, 1, 2, 3, and then summed all the ranks (`0+1+2+3=6`) on the process with rank 0, which finally printed all this output. The whole reduction operation is performed `n_iter` times, so that we get a more reproducible timing.
```

docs/test-suite/writing-portable-tests.md
```diff
@@ -0,0 +1,526 @@
...
+
+## Test requirements
+
+To be useful in the aforementioned scenarios, tests need to satisfy a number of requirements. 
+
+- Tests are implemented in the [ReFrame HPC testing framework](https://reframe-hpc.readthedocs.io/en/stable/).
+- Multiple test cases may be implemented for a single software package.
+- Tests should run in a reasonable amount of time (less than 1 hour) for all the scales for which it is defined to be valid.
```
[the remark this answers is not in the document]
casparvl: So... This is a great excuse to talk about it. Should it run in reasonable amount of time on a recent-generation HPC node? (i.e. 128 cores or so) Or also on a recent, powerful desktop, or decently sized VM (16 cores or so)?

Maybe it doesn't matter: if we say 'all scales', we also mean `1_core`. So then maybe we should say 'on a recent generation CPU / GPU'?

docs/test-suite/writing-portable-tests.md
```diff
@@ -0,0 +1,526 @@
...
+- There should be at least one light test case that can be run in less then ~5 minutes. This test should be marked with the 'CI' tag.
```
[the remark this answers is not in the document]
casparvl: Hmmmm, yes, although on a laptop, it might be too ambitious to require less than 5 minute runtime. I guess implicitely I mean: it should be 5 minutes on a `1_node` scale, given a decently-sized machine (16 cores or so, recent generation). I should probably make that _explicit_, but need to think how to do that in a short sentence :D

I like the addition of low-memory as well. I guess we should just say "on a modern day desktop / reasonably sized VM (in 2024 that means ~16 cores, 16 GB memory or so)."

docs/test-suite/writing-portable-tests.md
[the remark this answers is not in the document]
smoors: yeah "on a recent generation CPU / GPU" is fine to me

docs/test-suite/writing-portable-tests.md
```diff
@@ -0,0 +1,546 @@
...
+
+#### Requesting sufficient memory (mandatory)
+
+When developing the test, we don't know how much memory the node will have on which it will run. However, we _do_ know how much our application _needs_.
+
+We can declare this need using the `req_memory_per_node` hook. This hook is mandatory for all tests. If you are on a system with a scheduler that runs jobs within a cgroup and where you can use `mpirun` as the parallel launcher command in the ReFrame configuration, getting the memory consumption is easy. You can (temporarily) add a `postrun_cmds` the following to the class body of your test that extracts the maximum memory that was used within your cgroup. For cgroups v1, the syntax would be:
+
+```python
+   # Temporarily define postrun_cmds to make it easy to find out memory usage
+    postrun_cmds = ['MAX_MEM_IN_BYTES=$(</sys/fs/cgroup/memory/$(</proc/self/cpuset)/memory.max_usage_in_bytes)', 'echo "MAX_MEM_IN_MIB=$(($MAX_MEM_IN_BYTES/1048576))"']
```
smoors, line 328: i found out how to make it work for both mpirun and srun:
```suggestion
    postrun_cmds = ['MAX_MEM_IN_BYTES=$(</sys/fs/cgroup/memory/$(</proc/self/cpuset)/../memory.max_usage_in_bytes)', 'echo "MAX_MEM_IN_MIB=$(($MAX_MEM_IN_BYTES/1048576))"']
```
this gets the memory usage of the parent job, instead of the `batch` step, which is also slightly more accurate.
for cgroups v2, it will probably be similar but i cannot test that.

docs/test-suite/writing-portable-tests.md
```diff
@@ -0,0 +1,546 @@
...
+    ```bash
+    srun bash -c '<my_program> && MAX_MEM_IN_BYTES=$(</sys/fs/cgroup/memory/$(</proc/self/cpuset)/memory.max_usage_in_bytes) && echo "MAX_MEM_IN_MIB=$(($MAX_MEM_IN_BYTES/1048576))"'
+    ```
+    for cgroups v1, or
+    ```bash
+    srun bash -c '<my_program> && MAX_MEM_IN_BYTES=$(</sys/fs/cgroup/$(</proc/self/cpuset)/memory.peak) && echo "MAX_MEM_IN_MIB=$(($MAX_MEM_IN_BYTES/1048576))"'
+    ```
+    for cgroups v2, and then launch your batch job manually.
+    This way, the parallel launcher (`srun`) creates a new cgroup in which it will run _both_ your program, as well as check the maximum memory usage _in that same cgroup_.
```
smoors, line 388: with the change proposed above this can be removed :)

casparvl: Done in https://github.com/EESSI/docs/pull/262/commits/115810f48392390e8d9365d5889024378d2cc95a

docs/test-suite/writing-portable-tests.md
casparvl: Fixed in https://github.com/EESSI/docs/pull/262/commits/00c47fcf30ff724632d06eb4357cc2632beae328

smoors (approved):
lgtm

smoors (approved):

docs/test-suite/writing-portable-tests.md
```diff
@@ -0,0 +1,534 @@
...
+        """
+        Filtering function for single node tests
+        """
+        return [
+            k for (k, v) in SCALES.items()
+            if v['num_nodes'] == 1
+        ]
+       ...
+       scale = parameter(get_singlenode_scales())
+    ```
```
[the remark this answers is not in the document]
casparvl: Err, I'm not sure if that does what you think it does :) As far as I know (and I'm no expert btw), mkdocs interprets

\```python

as "here will be a section of python code". Whereas
\```
python
\```
Would be: here is a section of code, and the first line is `python` (and that's not what I intended to do).

docs/test-suite/writing-portable-tests.md
```diff
@@ -0,0 +1,534 @@
...
+[ RUN      ] EESSI_MPI4PY %module_name=mpi4py/3.1.5-gompi-2023b %scale=1_8_node /73046a73 @snellius:genoa+default
+[ RUN      ] EESSI_MPI4PY %module_name=mpi4py/3.1.5-gompi-2023b %scale=1cpn_4nodes /f08712a2 @snellius:genoa+default
+[ RUN      ] EESSI_MPI4PY %module_name=mpi4py/3.1.5-gompi-2023b %scale=1cpn_2nodes /23cd550b @snellius:genoa+default
+[ RUN      ] EESSI_MPI4PY %module_name=mpi4py/3.1.5-gompi-2023b %scale=4_cores /bb8e1349 @snellius:genoa+default
+[ RUN      ] EESSI_MPI4PY %module_name=mpi4py/3.1.5-gompi-2023b %scale=2_cores /4c0c7c9e @snellius:genoa+default
+[ RUN      ] EESSI_MPI4PY %module_name=mpi4py/3.1.5-gompi-2023b %scale=1_core /aa83ba9e @snellius:genoa+default
+
+...
+```
+on a system with 192 cores per node. I.e. any test of 2 nodes (384 cores) or above would be skipped because it exceeds our max task count.
```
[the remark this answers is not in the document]
casparvl: Excellent suggestion, this is something valuable I missed :)

docs/test-suite/writing-portable-tests.md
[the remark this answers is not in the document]
casparvl: Ah, no, the indentation is intentional: it should end up in the `!!! note` (and it does, by using that indentation)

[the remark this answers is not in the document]
Crivella: Oh ok.
It was about the indentation, but didn't realize it was intended, the python was an error on my part in trying to make the triple backtick work inside the github suggestion block

docs/test-suite/writing-portable-tests.md
```diff
@@ -0,0 +1,541 @@
...
+by 
+
+```python
+from eessi.testsuite.constants import SCALES
+...
+    # ReFrame will generate a test for each scale
+    scale = parameter(SCALES.keys())
+```
+
+the `SCALES` constant (TODO: API reference) contains a set of default scales at which we run all tests. For our `mpi4py` example, that is sufficient.
```
boegel, line 201: @casparvl Maybe remove the `TODO` for now, looks a bit awkward in the rendered docs, and we don't want to block this PR by not having API docs yet
```suggestion

the `SCALES` constant contains a set of default scales at which we run all tests. For our `mpi4py` example, that is sufficient.

File: docs/test-suite/writing-portable-tests.md [hunk 14]
```diff
@@ -0,0 +1,541 @@
...
+
+#### Replacing hard-coded module names (mandatory)
+
+If we write an `mpi4py` test, we typically want to run this for _all_ `mpi4py` modules that are currently on our `$MODULEPATH`. We do that by replacing
+
+```python
+    module_name = parameter(['mpi4py/3.1.4-gompi-2023a', 'mpi4py/3.1.5-gompi-2023b'])
+```
+
+by
```
boegel, line 258:
```suggestion
by using the `find_modules` utility function:

File: docs/test-suite/writing-portable-tests.md [hunk 15]
```diff
@@ -0,0 +1,541 @@
...
+
+```python
+    @run_after('init')
+    def set_modules(self):
+        hooks.set_modules(self)
+```
+
+This hook assumes that `self.module_name` has been set, but has the added advantage that a user running the EESSI test suite can overwrite the modules to load from the command line when running ReFrame (see [Overriding test parameters](https://www.eessi.io/docs/test-suite/usage/#overriding-test-parameters-advanced)).
```
boegel, line 282:
```suggestion
The `set_modules` hook assumes that `self.module_name` has been set, but has the added advantage that a user running the EESSI test suite can overwrite the modules to load from the command line when running ReFrame (see [Overriding test parameters](https://www.eessi.io/docs/test-suite/usage/#overriding-test-parameters-advanced)).

File: docs/test-suite/writing-portable-tests.md [hunk 16]
```diff
@@ -0,0 +1,541 @@
...
+
+#### Replacing hard-coded valid_systems (mandatory)
+
+The `valid_systems` attribute is a mandatory attribute to specify in a ReFrame test. However, we can set it to match any system:
+
+```python
+valid_systems = [*]
+```
+
+Normally, `valid_systems` is used as a way of guaranteeing that a system has the necessary properties to run the test. E.g. if we know that `my_gpu_system` has NVIDIA GPUs, and I have a test written for NVIDIA GPUs. I would specify `valid_systems['my_gpu_system']` for that test. This, however, is a surrogate for declaring what my test _needs_: I'm saying it needs `my_gpu_system`, while in fact I could make the more general statement 'this test needs NVIDIA GPUs'.
```
boegel, line 292:
```suggestion
Normally, `valid_systems` is used as a way of guaranteeing that a system has the necessary properties to run the test. For example, if we know that `my_gpu_system` has NVIDIA GPUs and I have a test written for NVIDIA GPUs, I would specify `valid_systems['my_gpu_system']` for that test. This, however, is a surrogate for declaring what my test _needs_: I'm saying it needs `my_gpu_system`, while in fact I could make the more general statement 'this test needs NVIDIA GPUs'.

File: docs/test-suite/writing-portable-tests.md [hunk 17]
```diff
@@ -0,0 +1,541 @@
...
+
+To keep the test system-agnostic we _can_ declare what the test needs by using ReFrame's concept of partition `features` (a string) and/or `extras` (a key-value pair) [see valid_systems documentation](https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html#reframe.core.pipeline.RegressionTest.valid_systems). E.g. a test could declare it _needs_ the `gpu` feature. Such a test will only be created by ReFrame for partitions that declare (in the ReFrame configuration file) that they have the `gpu` feature.
```
boegel, line 294:
```suggestion
To keep the test system-agnostic we _can_ declare what the test needs by using ReFrame's concept of partition `features` (a string) and/or `extras` (a key-value pair); [see the ReFrame documentation on `valid_systems`](https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html#reframe.core.pipeline.RegressionTest.valid_systems). For example, a test could declare it _needs_ the `gpu` feature. Such a test will only be created by ReFrame for partitions that declare (in the ReFrame configuration file) that they have the `gpu` feature.

File: docs/test-suite/writing-portable-tests.md [hunk 18]
```diff
@@ -0,0 +1,541 @@
...
+...
+'features': [
+     FEATURES[GPU],
+],
+'extras': {
+    GPU_VENDOR: GPU_VENDORS[NVIDIA],
+},
+```
+
+In practice, one will rarely hard-code this `valid_systems` string. Instead, we have a hook [filter_valid_systems_by_device_type](TODO API REF). It does the above, and a bit more: it also checks if the module that the test is generated for is CUDA-enabled (in case of a test for `NVIDIA` GPUs), and _only then_ will it generate a GPU-based test. Calling this hook is mandatory for all tests (even if just to declare they need a CPU to run).
```
boegel, line 317:
```suggestion
In practice, one will rarely hard-code this `valid_systems` string. Instead, we have a hook `filter_valid_systems_by_device_type`. It does the above, and a bit more: it also checks if the module that the test is generated for is CUDA-enabled (in case of a test for `NVIDIA` GPUs), and _only then_ will it generate a GPU-based test. Calling this hook is mandatory for all tests (even if just to declare they need a CPU to run).

File: docs/test-suite/writing-portable-tests.md [hunk 19]
```diff
@@ -0,0 +1,541 @@
...
+
+Another aspect is that not all ReFrame partitions may be able to run tests of all of the standard `SCALES`. Each ReFrame partition must add the subset of `SCALES` it supports to its list of features.  A test case can declare it needs a certain scale. For example, a test case using the `16_nodes` scale needs a partition with at least 16 nodes. The [filter_supported_scales](TODO API REF) hook then filters out all partitions that do not support running jobs on 16 nodes. Calling this hook is also mandatory for all tests.
+
+There may be other hooks that facilitate valid system selection for your tests, but please check the [API documentation](TODO: INSERT REFERENCE TO API DOCS) for a full list.
```
boegel, line 321:
```suggestion
There may be other hooks that facilitate valid system selection for your tests, but please check the [API documentation](https://github.com/EESSI/test-suite/blob/main/eessi/testsuite/hooks.py) for a full list.

File: docs/test-suite/writing-portable-tests.md [hunk 20]
```diff
@@ -0,0 +1,541 @@
...
+
+```python
+@run_after('setup')
+def set_binding(self):
+    hooks.set_compact_process_binding(self)
+```
+
+For pure MPI codes, it will bind rank 0 to core 0, rank 1 to core 1, etc. For hybrid codes (MPI + OpenMP, or otherwise codes that do both multiprocessing and multithreading at the same time), it will bind to consecuitive sets of cores. E.g. if a single process uses 4 cores, it will bind rank 0 to cores 0-3, rank 1 to cores 4-7, etc. 
+
+To impose this binding, the hook sets environment variables that should be respected by the parallel launcher used to launch your application. Check the [TODO: API Documentation] to see which parallel launchers are currently supported. The use of this hook is optional, but generally recommended for all multiprocessing codes.
```
boegel, line 416:
```suggestion
To impose this binding, the hook sets environment variables that should be respected by the parallel launcher used to launch your application. Check the [code](https://github.com/EESSI/test-suite/blob/main/eessi/testsuite/hooks.py) to see which parallel launchers are currently supported. The use of this hook is optional, but generally recommended for all multiprocessing codes.
```

docs/test-suite/writing-portable-tests.md
```diff
@@ -0,0 +1,541 @@
...
+
+For multithreading codes, there `set_compact_thread_binding` hook is an equivalent hook that can do thread binding, if supported multithreading frameworks are used (e.g. Intel or GNU OpenMP, see the [TODO API documentation] for all supported frameworks):
```
boegel, line 418:
```suggestion
For multithreading codes, there `set_compact_thread_binding` hook is an equivalent hook that can do thread binding, if supported multithreading frameworks are used (e.g. Intel or GNU OpenMP, see the [code](https://github.com/EESSI/test-suite/blob/main/eessi/testsuite/hooks.py) for all supported frameworks):
```

docs/test-suite/writing-portable-tests.md
```diff
@@ -0,0 +1,541 @@
...
+@run_after('setup')
+def set_binding(self):
+    hooks.set_compact_thread_binding(self)
+```
+
+The use of this hook is optional but recommended in most cases. Note that thread binding can sometimes cause unwanted behaviour: even if e.g. 8 cores are allocated to the process and 8 threads are launched, we have seen codes that bind all those threads to a single core (e.g. core 0) when core binding is enabled. Please verify that enabling core binding does not introduce any unwanted binding behaviour for your code.
+
+#### Defining OMP_NUM_THREADS (recommended)
+
+The `set_omp_num_threads` hook sets the `OMP_NUM_THREADS` environment variable based on the number of `cpus_per_task` defined in the ReFrame test (which in turn is typically set by the `assign_tasks_per_compute_unit` hook). For OpenMP codes, it is generally recommended to call this hook, to ensure they launch the correct amount of threads.
```
boegel, line 430:
```suggestion
The `set_omp_num_threads` hook sets the `$OMP_NUM_THREADS` environment variable based on the number of `cpus_per_task` defined in the ReFrame test (which in turn is typically set by the `assign_tasks_per_compute_unit` hook). For OpenMP codes, it is generally recommended to call this hook, to ensure they launch the correct amount of threads.

File: docs/test-suite/writing-portable-tests.md [hunk 23]
```diff
@@ -0,0 +1,541 @@
...
+
+#### Skipping tests instances when required (optional)
+Preferably, we prevent test instances from being generated (i.e. before ReFrame's `setup` phase) if we know that they cannot run on a certain system. However, sometimes we need information on the nodes that will run it, which is only available _after_ the `setup` phase. That is the case for anything where we need information from e.g. the [reframe.core.pipeline.RegressionTest.current_partition](https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html#reframe.core.pipeline.RegressionTest.current_partition). The `assign_tasks_per_compute_unit` hook for example uses this property to get the core count of a node, and thus needs to be executed after the `setup` phase.
+
+For example, we might know that a test only scales to around 300 tasks, and above that, execution time increases rapidly. In that case, we'd want to skip any test instance that results in a larger amount of tasks, but we only know this after `assign_tasks_per_compute_unit` has been called. E.g. the `2_nodes` scale would run fine on systems with 128 cores per node, but would exceed the task limit of 300 on systems with `192` cores per node.
```
boegel, line 435:
```suggestion
For example, we might know that a test only scales to around 300 tasks, and above that, execution time increases rapidly. In that case, we'd want to skip any test instance that results in a larger amount of tasks, but we only know this after `assign_tasks_per_compute_unit` has been called. For example, the `2_nodes` scale would run fine on systems with 128 cores per node, but would exceed the task limit of 300 on systems with `192` cores per node.

File: docs/test-suite/writing-portable-tests.md [hunk 24]
```diff
@@ -0,0 +1,541 @@
...
+
+We can skip any generated test cases using the `skip_if` function. E.g. to skip the test if the total task count exceeds 300, we'd need to call `skip_if` _after_ the task count has been set by `assign_tasks_per_compute_unit`:
```
boegel, line 437:
```suggestion
We can skip any generated test cases using the `skip_if` function. For example, to skip the test if the total task count exceeds 300, we'd need to call `skip_if` _after_ the task count has been set by `assign_tasks_per_compute_unit`:

File: docs/test-suite/writing-portable-tests.md [hunk 25]
```diff
@@ -0,0 +1,541 @@
...
+[ RUN      ] EESSI_MPI4PY %module_name=mpi4py/3.1.5-gompi-2023b %scale=2_cores /4c0c7c9e @snellius:genoa+default
+[ RUN      ] EESSI_MPI4PY %module_name=mpi4py/3.1.5-gompi-2023b %scale=1_core /aa83ba9e @snellius:genoa+default
+
+...
+```
+on a system with 192 cores per node. I.e. any test of 2 nodes (384 cores) or above would be skipped because it exceeds our max task count.
+
+**NOTE**: The order in which [ReFrame pipeline hooks](https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html#pipeline-hooks) (methods decorated with `run_after` or `run_before`) are called is the same in which they are attached/defined.
+This is important in case we want to call hooks for the same stage (`init`/`setup`/...) in different functions (for cleanliness of the code or any other reason).
+For example, any pipeline hook attached to the `setup` step making use of `self.num_tasks`, should be defined after the function calling the test-suite hook `assign_tasks_per_compute_unit`
```
boegel, line 474: Make this a proper note:
```suggestion
!!! note
    The order in which [ReFrame pipeline hooks](https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html#pipeline-hooks) (methods decorated with `run_after` or `run_before`) are called is the same in which they are attached/defined.

    This is important in case we want to call hooks for the same stage (`init`/`setup`/...) in different functions (for cleanliness of the code or any other reason).

    For example, any pipeline hook attached to the `setup` step making use of `self.num_tasks`, should be defined after the function calling the test-suite hook `assign_tasks_per_compute_unit`.

**Review by boegel — CHANGES_REQUESTED** [review]

File: docs/test-suite/writing-portable-tests.md (hunk 13 above)
Comment (casparvl) [reply docs/test-suite/writing-portable-tests.md:201]: True, they were there for my own sake. Can we leave a comment that doesn't get rendered? That would be helpful to our future selves...

Comment (boegel) [reply docs/test-suite/writing-portable-tests.md:201]: You could use HTML-style comments, like:

```
<!-- TODO add API reference -->
```

but I'm not sure it's worth it. Once we have the API docs, we'll need to go through this whole page again, I think, since the TODO didn't seem consistent to me (not all hooks, etc.)

casparvl: Fair enough, I now replaced everything with references to code :)

boegel (approved):
lgtm, thanks a lot for all the effort on this @casparvl, this is a great piece of documentation!

PR #263, opened 2024-08-14
update `pip install` instructions in README

casparvl:


README.md
```diff
@@ -16,7 +16,7 @@ which makes it very easy to preview the result of the changes you make locally.
 
 * First, install ``mkdocs``:
 
-        pip install mkdocs-material mkdocs-git-revision-date-localized-plugin
+        pip install mkdocs-material mkdocs-git-revision-date-localized-plugin mkdocs-macros-plugin mkdocs-redirects
```
boegel, line 19: We should use `pip install -r requirements.txt` instead, and make sure `requirements.txt` is correct?

boegel (changes requested):

README.md
casparvl: Lol, why do I always get more work when I try to be nice and fix something ;-) (but you are right, of course)

casparvl: ok, `requirements.txt` _was_ actually up to date. I just updated this instruction to `pip install -r requirements.txt`, and all should be right in the world.

boegel (approved):

PR #267, opened 2024-08-21
run workflow to update overview of available software only twice a day

boegel:
Running every 4 hours too frequently results in multiple PRs like #265 being opened that than need to be closed because another more recent PR with the same updates got merged, so let's tune it down a bit...


trz42 (approved):
makes sense

PR #316, opened 2024-10-07
Updated existing docs on how to build on top of EESSI to use `EESSI-extend`

casparvl:
This now depends on 
- [x] https://github.com/EESSI/software-layer/pull/778


docs/using_eessi/building_on_eessi.md
```diff
@@ -6,67 +6,113 @@ Building on top of EESSI with EasyBuild is relatively straightforward. One cruci
 ### Starting the EESSI software environment
 Start your environment as described [here](../using_eessi/setting_up_environment.md)
 
-### Configure EasyBuild
-To configure EasyBuild, first, check out the [EESSI software-layer repository](https://github.com/EESSI/software-layer.git). We advise you to check out the branch corresponding to the version of EESSI you would like to use.
+### Using the EESSI-extend module
+The `EESSI-extend` module facilitates building on top of EESSI using EasyBuild. It does a few key things:
 
-If you are unsure which version you are using, you can run
+1. It configures EasyBuild to match how the rest of the EESSI software is build
```
ocaisa, line 12:
```suggestion
1. It configures EasyBuild to match how the rest of the EESSI software is built
```

docs/using_eessi/building_on_eessi.md
```diff
@@ -6,67 +6,113 @@ Building on top of EESSI with EasyBuild is relatively straightforward. One cruci
...
+ - EESSI_SITE_INSTALL is either defined or not and cannot be used with another environment variable. A site installation is done in a defined location and any installations there are (by default) world readable. Set this environmnet variable if you are a site hosting a system that has EESSI available, and you want to build on top to make additional modules available to all of your users.
+ - EESSI_CVMFS_INSTALL is either defined or not and cannot be used with another environment variable. A CVMFS installation targets a defined location which will be ingested into CVMFS and is only useful for CVMFS administrators. Set this environment variable if you manage your own CVMFS repository where the software has to be built on top of EESSI.
```
ocaisa, line 26: This is true, but it is not really of relevance here...

docs/using_eessi/building_on_eessi.md
```diff
@@ -6,67 +6,113 @@ Building on top of EESSI with EasyBuild is relatively straightforward. One cruci
...
+The `EESSI-extend` module recognizes a few environment variables. To print an up-to-date list, check the module itself
 ```
-echo ${EESSI_VERSION}
+module help EESSI-extend/2023.06-easybuild
 ```
-to check it.
 
-To build on top of e.g. version `2023.06` of the EESSI software stack, we check it out, and go into that directory:
+The key environment variables it will respect are:
+ - EESSI_USER_INSTALL can be set to a location to install modules for use by the user only. The location must already exist on the filesystem. Set this environment variable if you're an end-user who wants to install additional software just for yourself.
```
ocaisa, line 23:
```suggestion
 - `EESSI_USER_INSTALL` can be set to a location to install modules for use by the user only. The location must already exist on the filesystem. Set this environment variable if you're an end-user who wants to install additional software just for yourself.
```

docs/using_eessi/building_on_eessi.md
```diff
@@ -6,67 +6,113 @@ Building on top of EESSI with EasyBuild is relatively straightforward. One cruci
...
+ - EESSI_PROJECT_INSTALL can be set to a location to install modules for use by a project. The location must already exist on the filesystem and you should ensure that the location has the correct Linux group and the SGID permission is set on that directory (`chmod g+s $EESSI_PROJECT_INSTALL`) so that all members of the group have permission to read and write installations. Set this environment variable if you're an end-user who's part of a project, and wants to create additional installations that should be available to all project members.
```
ocaisa, line 24:
```suggestion
 - `EESSI_PROJECT_INSTALL` can be set to a location to install modules for use by a project. The location must already exist on the filesystem and you should ensure that the location has the correct Linux group and the SGID permission is set on that directory (`chmod g+s $EESSI_PROJECT_INSTALL`) so that all members of the group have permission to read and write installations. Set this environment variable if you're an end-user who's part of a project, and wants to create additional installations that should be available to all project members.
```

docs/using_eessi/building_on_eessi.md
```diff
@@ -6,67 +6,113 @@ Building on top of EESSI with EasyBuild is relatively straightforward. One cruci
...
+ - EESSI_SITE_INSTALL is either defined or not and cannot be used with another environment variable. A site installation is done in a defined location and any installations there are (by default) world readable. Set this environmnet variable if you are a site hosting a system that has EESSI available, and you want to build on top to make additional modules available to all of your users.
```
ocaisa, line 25:
```suggestion
 - `EESSI_SITE_INSTALL` is either defined or not and cannot be used with another environment variable. A site installation is done in a defined location and any installations there are (by default) world readable. Set this environment variable if you are a site hosting a system that has EESSI available, and you want to build on top to make additional modules available to all of your users.
```

docs/using_eessi/building_on_eessi.md
```diff
@@ -6,67 +6,113 @@ Building on top of EESSI with EasyBuild is relatively straightforward. One cruci
...
+ - EESSI_CVMFS_INSTALL is either defined or not and cannot be used with another environment variable. A CVMFS installation targets a defined location which will be ingested into CVMFS and is only useful for CVMFS administrators. Set this environment variable if you manage your own CVMFS repository where the software has to be built on top of EESSI.
+
+Note that if none of the environment variables above are defined, an EESSI_USER_INSTALL is assumed with a value of $HOME/EESSI.
```
ocaisa, line 28:
```suggestion
Note that if none of the environment variables above are defined, an `EESSI_USER_INSTALL` is assumed with a value of `$HOME/EESSI`.
```

docs/using_eessi/building_on_eessi.md
```diff
@@ -6,67 +6,113 @@ Building on top of EESSI with EasyBuild is relatively straightforward. One cruci
...
-module load netCDF/4.9.0-gompi-2022a
+# Loading EESSI-extend makes the netcdf4-python module available on the MODULEPATH
+module load EESSI-extend/2023.06-easybuild
+module load netcdf4-python/1.6.5-foss-2023b
 ```
 
+in your job script.
+
+!!! Note
+    The `EESSI-extend` module does _not_ add `EESSI_SITE_INSTALL` to the `MODULEPATH`. As a system administrator, you will have to make sure that's on the `MODULEPATH` of your users through other means if you want your users to be able to use module installed by `EESSI-extend`.
```
ocaisa, line 114: This is not really correct, EESSI initialisation by default adds this location to `MODULEPATH` so if you are using our init script or our modules this will automatically be in your `MODULEPATH`

ocaisa (changes requested):

docs/using_eessi/building_on_eessi.md
casparvl: Ehm, what's the 'this' here exactly? I'm assuming it's relevant to explain the environment variables that can be set here, no?

ocaisa: `EESSI_CVMFS_INSTALL`, it can only really work for us (who can configure write access to that location)

docs/using_eessi/building_on_eessi.md
casparvl: Ah, great. But only if `EESSI_SITE_INSTALL` is set when initializing, I suppose? I'll change the note to that  - make sure that `EESSI_SITE_INSTALL` is set before users initialize EESSI.

ocaisa: No, it is always set (but of course usually empty)

docs/using_eessi/building_on_eessi.md
casparvl: What if I host my own Stratum 0 at `software.surf.nl`, and use a (the) build container to mount that with a writeable overlay. Can I then build _on top of EESSI_ in my own CVMFS repo?

docs/using_eessi/building_on_eessi.md
ocaisa: Yes, this is all correct.

ocaisa: and `EESSI_USER_INSTALL` is the default with a location of `$HOME/EESSI`

docs/using_eessi/building_on_eessi.md
ocaisa: Yes, you can do that today by using `EESSI_PROJECT_INSTALL` but that won't give you world readable rights on the installations.

Other scenarios could be supported, we'd just need the specific use case and then implement the logic. For what you mention, you really just want `EESSI_PROJECT_INSTALL` but with world readable permissions.

docs/using_eessi/building_on_eessi.md
casparvl: Ok, I rewrote these sections, I think it should be ok now.

docs/using_eessi/building_on_eessi.md
casparvl: Ok, well, something for later, we're note that far yet anyway :)

docs/using_eessi/building_on_eessi.md
```diff
@@ -6,67 +6,109 @@ Building on top of EESSI with EasyBuild is relatively straightforward. One cruci
...
+The `EESSI-extend` module recognizes a few environment variables. To print an up-to-date list, check the module itself
 ```
-echo ${EESSI_VERSION}
+module help EESSI-extend/2023.06-easybuild
 ```
-to check it.
 
-To build on top of e.g. version `2023.06` of the EESSI software stack, we check it out, and go into that directory:
+The installation prefix is determined by `EESSI-extend` through the following logic:
+1. If `EESSI_CVMFS_INSTALL` is set, software inst installed in `EESSI_SOFTWARE_PATH`. This variable shouldn't be used by users and would only be used by CVMFS administrators of the EESSI repository.
```
ocaisa, line 23:
```suggestion
1. If `EESSI_CVMFS_INSTALL` is set, software inst installed in `EESSI_SOFTWARE_PATH`. This variable shouldn't be used by users and would only be used by CVMFS administrators of the EESSI repository.
```
```suggestion
1. If `EESSI_CVMFS_INSTALL` is set, software is installed in `EESSI_SOFTWARE_PATH`. This variable shouldn't be used by users and would only be used by CVMFS administrators of the EESSI repository.
```

docs/using_eessi/building_on_eessi.md
```diff
@@ -6,67 +6,109 @@ Building on top of EESSI with EasyBuild is relatively straightforward. One cruci
...
+2. If `EESSI_SITE_INSTALL` is set (and this is set by default when the EESSI environment is initialized) and the EESSI site installation prefix (`/cvmfs/software.eessi.io/host_injections/...`) is writeable, this prefix will be used. This is typically where sites hosting a system that has EESSI deployed would install additional software on top of EESSI and make it available to all their users.
```
ocaisa, line 24:
```suggestion
2. If `EESSI_SITE_INSTALL` is set and the EESSI site installation prefix (`/cvmfs/software.eessi.io/host_injections/...`) is writeable, this prefix will be used. This is typically where sites hosting a system that has EESSI deployed would install additional software on top of EESSI and make it available to all their users.
```
This is not set by default, what is done by default is _exposing_ the modules that are installed with this approach.

ocaisa (commented):
One note would be that `EESSI_PROJECT_INSTALL` and `EESSI_USER_INSTALL` are set to a path, and this is the path where installations are made. With this approach you can have multiple projects with different shared software stacks

docs/using_eessi/building_on_eessi.md
casparvl: Ah, I figured that was also a `PATH`-like variable. It isn't. Looking at the `EESSI-extend` module:

```
elseif (os.getenv("EESSI_SITE_INSTALL") ~= nil) then
  -- Make sure no other EESSI install environment variables are set
  if ((os.getenv("EESSI_PROJECT_INSTALL") ~= nil) or (os.getenv("EESSI_USER_INSTALL") ~= nil)) then
    LmodError("You cannot use EESSI_SITE_INSTALL in combination with any other EESSI_*_INSTALL environment variables")
  end
  easybuild_installpath = string.gsub(os.getenv("EESSI_SOFTWARE_PATH"), 'versions', 'host_injections')
```

Not really a docs question, but it's a bit awkward that this path is recalculated when the EESSI modulefile already sets `EESSI_SITE_MODULEPATH`. Admittedly, that is _not_ the installpath (that would be 2 dirs up).

Ideally, I would _also_ set `EESSI_SITE_INSTALLPATH` in the EESSI modulefile (and make `EESSI_SITE_MODULEPATH relative to that). But that's a different story.

The point is, right now, it's very hard to explain to a user _where_ this site installation will end up. We don't have an environment variable pointing to that prefix :\ For now, that `...` is the best I can do, but it's very a-specific.

casparvl:
This now depends on https://github.com/EESSI/software-layer/pull/778

boegel:
> This now depends on [EESSI/software-layer#778](https://github.com/EESSI/software-layer/pull/778)

Better note that in the PR description...

docs/using_eessi/building_on_eessi.md
```diff
@@ -6,67 +6,109 @@ Building on top of EESSI with EasyBuild is relatively straightforward. One cruci
...
 ```
-echo ${EESSI_VERSION}
+module help EESSI-extend/2023.06-easybuild
 ```
-to check it.
 
-To build on top of e.g. version `2023.06` of the EESSI software stack, we check it out, and go into that directory:
+The installation prefix is determined by `EESSI-extend` through the following logic:
+1. If `$EESSI_CVMFS_INSTALL` is set, software is installed in `$EESSI_SOFTWARE_PATH`. This variable shouldn't be used by users and would only be used by CVMFS administrators of the EESSI repository.
+2. If `$EESSI_SITE_INSTALL` is set, the EESSI site installation prefix (`$EESSI_SITE_INSTALLPATH`) will be used. This is typically where sites hosting a system that has EESSI deployed would install additional software on top of EESSI and make it available to all their users.
```
ocaisa, line 24:
```suggestion
2. If `$EESSI_SITE_INSTALL` is set, the EESSI site installation prefix (`$EESSI_SITE_SOFTWARE_PATH`) will be used. This is typically where sites hosting a system that has EESSI deployed would install additional software on top of EESSI and make it available to all their users.
```
```suggestion
2. If `$EESSI_SITE_INSTALL` is set, the EESSI site installation prefix (`$EESSI_SITE_INSTALLPATH`) will be used. This is typically where sites hosting a system that has EESSI deployed would install additional software on top of EESSI and make it available to all their users.
```

docs/using_eessi/building_on_eessi.md
```diff
@@ -6,67 +6,109 @@ Building on top of EESSI with EasyBuild is relatively straightforward. One cruci
...
+3. If `$EESSI_PROJECT_INSTALL` is set (and `$EESSI_USER_INSTALL` is not set), this prefix will be used. You should use this if you want to install additional software on top of EESSI that should also be usable by your project partners on the same system.
```
ocaisa, line 25: Should give an example here to make it clear you provide a path

docs/using_eessi/building_on_eessi.md
```diff
@@ -6,67 +6,109 @@ Building on top of EESSI with EasyBuild is relatively straightforward. One cruci
...
+4. If `$EESSI_USER_INSTALL` is set, this prefix will be used. You should use this if yo uwant to install additional software on top of EESSI just for your own user.
```
ocaisa, line 26:
```suggestion
4. If `$EESSI_USER_INSTALL` is set, this prefix will be used. You should use this if you want to install additional software on top of EESSI just for your own user.
```

ocaisa: An example would again help here

ocaisa (changes requested):

docs/using_eessi/building_on_eessi.md
casparvl: Example added, for both user and project installs.

ocaisa (approved):

PR #317, opened 2024-10-08
Update existing Doc on how to use EESSI

TopRichard:
Updating `setting_up_environment.md` in EESSI docs by adding EESSI module usage


docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,21 @@
 # Setting up your environment
 
-To set up the EESSI environment, simply run the command:
+To set up the EESSI environment, either use EESSI with Lmod:
```
ocaisa, line 3: I think you need to give people a little more guidance here about when to use which approach. If you are already using Lmod >= X, you can use the first option (and you should consider unsetting `MODULEPATH`). If you use anything other than Lmod you should definitely unset `MODULEPATH` as if Lmod sees it it will try to build a cache for it. If you use Tmod or have no module tool, then you need to initialise Lmod.

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,48 @@
 # Setting up your environment
 
-To set up the EESSI environment, simply run the command:
+To set up the EESSI environment, you can either use EESSI with Lmod or use a bash initialisation script.
```
trz42, line 3: This sounds a bit strange to me "set up the EESSI environment ... by ... _use_ EESSI with Lmod or _use_ a bash ..."

Maybe use different verbs and objects? For example
```suggestion
To set up the EESSI environment, you can either load an EESSI module with Lmod or source a bash initialisation script.
```

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,48 @@
...
+
+## Using EESSI with Lmod
```
trz42, line 5: Adjust to what is used above, for example,
```suggestion
## Loading an EESSI module with Lmod
```

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,48 @@
...
+
+There are a few different scenarios where you may want to use EESSI with Lmod:
```
trz42, line 7: Same here
```suggestion
There are a few different scenarios where you may want to set up the EESSI environment by loading an EESSI module with Lmod:
```

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,48 @@
...
+
+1.  You already using Lmod version >= 8.6
```
trz42, line 9: Nitpicking

```suggestion
1.  You are already using Lmod with version >= 8.6
```

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,48 @@
...
+
+    In this case you should consider unsetting the `MODULEPATH`, as you do not really want to mix modules coming from EESSI and from your system
```
trz42, line 11: Maybe?
```suggestion
    In this case, we recommend unsetting the `MODULEPATH`, because EESSI is not designed to mix modules coming from EESSI and from your system.
```

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,48 @@
...
+    since they are effectively two different OSes (though this may not be possible at some sites, and some compromise may be necessary):
```
trz42, line 12:
```suggestion
    The underlying reason is that very likely EESSI and your system are based on two different operating system distributions - EESSI uses Gentoo, your system likely uses some other distribution. Note, unsetting `MODULEPATH` may not be possible at some sites, and some compromise may be necessary:
```

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,48 @@
...
+
+    ``` { .bash .copy }
+    unset MODULEPATH
+    export MODULEPATH=/cvmfs/software.eessi.io/init/modules
+    module load EESSI/2023.06
+    ```
+
+    :clap: Your environment is now set up, you are ready to start running software provided by EESSI!
+
+2.  You are using an older version of Lmod or any other tool utilizing `MODULEPATH` (Tmod, etc.)
```
trz42, line 22:
```suggestion
2.  You are using an Lmod with a version older than 8.6 or any other tool utilizing `MODULEPATH` (Tmod, etc.)
```

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,48 @@
...
+    modules: 
+
+    ``` { .bash .copy }
+    unset MODULEPATH
+    source /cvmfs/software.eessi.io/versions/2023.06/init/lmod/bash
+    ```
+
+    :clap: Your environment is now set up, you are ready to start running software provided by EESSI!
+
+3.  Should Lmod be unavailable and `MODULEPATH` not utilized, you can use EESSI as a module by directly sourcing the Lmod initialization script:
```
trz42, line 35:
```suggestion
3.  Should Lmod be unavailable and `MODULEPATH` not utilized, you can initialise EESSI via an Lmod module by directly sourcing the Lmod initialisation script (this script automatically loads the EESSI module):
```

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,48 @@
...
+
+    ``` { .bash .copy }
+    source /cvmfs/software.eessi.io/versions/2023.06/init/lmod/bash
+    ```
+
+    :clap: Your environment is now set up, you are ready to start running software provided by EESSI!
+
+## Use the EESSI `bash` initialisation script
```
trz42, line 43:
```suggestion
## Sourcing the EESSI `bash` initialisation script
```

trz42 (changes requested):
Looks good ... it feels a bit like one would have to know a bit about the system you're running on or read through several alternatives before you get going. Too many choices. Would be nice if one could just say `eessi init` and the best option is chosen. Then if one wants more control one specifies options to this command (or uses one of the alternatives listed here).

ocaisa:
It is perhaps even more complicated than this, as we haven't told them what to do if they are using a shell other than `bash`, for Lmod there are multiple permitted options.

trz42:
> It is perhaps even more complicated than this, as we haven't told them what to do if they are using a shell other than `bash`, for Lmod there are multiple permitted options.

That's right. I wonder if - eventually - we can come up with a single command that works for most users.

For now, it's ok to accept that such a single command doesn't exist yet and we provide information about all the alternatives.

ocaisa:
You can, but that will ultimately be a wrapper around Lmod (you're basically trying to reimplement the same thing that `module load ...` does), and then you would still need shell specific versions of the wrappers.

Unless such a command starts a subshell...which is possible.

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,48 @@
...
-To set up the EESSI environment, simply run the command:
+To set up the EESSI environment, you can either load an EESSI module with Lmod or source a bash initialisation script.
+
+## Loading an EESSI module with Lmod
+
+There are a few different scenarios where you may want to set up the EESSI environment by loading an EESSI module with Lmod:
+
+1.  You are already using Lmod with version >= 8.6
+
+    In this case, we recommend unsetting the `MODULEPATH`, because EESSI is not designed to mix modules coming from EESSI and from your system.
```
ocaisa, line 11:
```suggestion
    In this case, we _recommend_ unsetting the `MODULEPATH`, because EESSI is not designed to mix modules coming from EESSI and from your system.
```

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,48 @@
...
+    The underlying reason is that very likely EESSI and your system are based on two different operating system distributions - EESSI uses Gentoo, your system likely uses some other distribution. Note, unsetting `MODULEPATH` may not be possible at some sites, and some compromise may be necessary:
```
ocaisa, line 12:
```suggestion 
```

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,48 @@
...
+
+    ``` { .bash .copy }
+    unset MODULEPATH
+    export MODULEPATH=/cvmfs/software.eessi.io/init/modules
+    module load EESSI/2023.06
+    ```
+
```
ocaisa, line 19:
```suggestion
    !!! note "Why should I unset `MODULEPATH`?"
        Unsetting `MODULEPATH` may not be possible at some sites, and some compromise or workaround may
        be necessary. The underlying reason to suggest this is that EESSI and your system are most likely based on
        two different operating system distributions - EESSI uses Gentoo Prefix, your system almost certainly uses some other
        distribution. If you can find a way to ensure that the _software stacks_ from your site and EESSI do not mix
        (in particular when someone is building new software!), then this should be good enough.

```

ocaisa (changes requested):
Couple of other suggestions

docs/using_eessi/setting_up_environment.md
ocaisa: It may be better to move this up entirely as the reasoning is value for all items

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,48 @@
...
+    modules: 
+
+    ``` { .bash .copy }
+    unset MODULEPATH
+    source /cvmfs/software.eessi.io/versions/2023.06/init/lmod/bash
+    ```
+
+    :clap: Your environment is now set up, you are ready to start running software provided by EESSI!
+
+3.  Should Lmod be unavailable and `MODULEPATH` not utilized, you can initialise EESSI via an Lmod module by directly sourcing the Lmod initialisation script (this script automatically loads the EESSI module):
```
ocaisa, line 35: To keep things simple, we should probably start with this and then have a collapsed section for "Other scenarios" that mentions the two other approaches (and the note/warning about unsetting the `MODULEPATH`). That way the two approaches are equivalent for typical use case.

ocaisa: See https://squidfunk.github.io/mkdocs-material/reference/admonitions/?h=admon#collapsible-blocks

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,48 @@
...
+
+    ``` { .bash .copy }
+    source /cvmfs/software.eessi.io/versions/2023.06/init/lmod/bash
```
ocaisa, line 38: We should make it clear that there are a range of options available here

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,48 @@
...
+    ```
+
+    :clap: Your environment is now set up, you are ready to start running software provided by EESSI!
+
+## Sourcing the EESSI `bash` initialisation script
+
+You can initialise EESSI (in a non-reversible way) by running the command:
```
ocaisa, line 45: I think it is worth explicitly saying that this command only works with bash, and you use another shell you should take the other approach.

docs/using_eessi/setting_up_environment.md
TopRichard: > It may be better to move this up entirely as the reasoning is value for all items

Moved it upwards

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,55 @@
...
-To set up the EESSI environment, simply run the command:
+To set up the EESSI environment, you can either load an EESSI module with Lmod or source a bash initialisation script.
+
+## Loading an EESSI module with Lmod
+
+There are a few different scenarios where you may want to set up the EESSI environment by loading an EESSI module with Lmod:
+
+   !!! note "Why do we recommend to unset `MODULEPATH`?"
+
+        Unsetting MODULEPATH may not be possible at some sites, and some compromise or workaround may
```
boegel, line 11:
```suggestion
        Unsetting the `$MODULEPATH` environment variable, which tells Lmod in which directories environment module files are available, may not be possible at some sites, and some compromise or workaround may
```

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,55 @@
...
+        be necessary. The underlying reason to suggest this is that EESSI and your system are most likely based on
+        two different operating system distributions - EESSI uses Gentoo Prefix, your system almost certainly uses some other
```
boegel, line 13: I wouldn't mention Gentoo Prefix here, keep it more general (compatibility layer + link to the corresponding page)

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,55 @@
...
+        distribution. If you can find a way to ensure that the _software stacks_ from your site and EESSI do not mix
```
boegel, line 14: `distribution` -> `Linux distribution`

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,55 @@
...
+        (in particular when someone is building new software!), then this should be good enough.
+
+1.  You are already using Lmod with version >= 8.6
+
+    In this case, we _recommend_ unsetting the `MODULEPATH`, because EESSI is not designed to mix modules coming from EESSI and from your system.
```
boegel, line 19:
```
    In this case, we _recommend_ unsetting `$MODULEPATH`, because EESSI is not designed to mix modules coming from EESSI and from your system.
```

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,55 @@
...
+
+    ``` { .bash .copy }
+    unset MODULEPATH
+    export MODULEPATH=/cvmfs/software.eessi.io/init/modules
```
boegel, line 23: If you're doing a hard set, then there's no need to `unset`?

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,55 @@
...
+    module load EESSI/2023.06
+    ```
+
+    :clap: Your environment is now set up, you are ready to start running software provided by EESSI!
+
+2.  You are using an Lmod with a version older than 8.6 or any other tool utilizing `MODULEPATH` (Tmod, etc.)
```
boegel, line 29: `Tmod` is not the official name, that's sort of "slang", use "Tcl-based Environment Modules" + link that to https://modules.sourceforge.net/

docs/using_eessi/setting_up_environment.md
```diff
@@ -1,6 +1,55 @@
...
+
+    You should unset MODULEPATH to prevent Lmod from attempting to build a cache for your module tree (as this can be very slow if you have
```
boegel, line 31:
```suggestion
    You should unset `$MODULEPATH` to prevent Lmod from attempting to build a cache for your module tree (as this can be very slow if you have

File: docs/using_eessi/setting_up_environment.md [hunk 24]
```diff
@@ -1,6 +1,55 @@
...
+    a lot of modules). Again, unsetting the MODULEPATH should be considered as a good idea in general so you do not mix local and EESSI
```
boegel, line 32:
```suggestion
    a lot of modules). Again, unsetting the `$MODULEPATH` should be considered as a good idea in general so you do not mix local and EESSI

File: docs/using_eessi/setting_up_environment.md (hunk 14 above)
Comment (boegel) [reply docs/using_eessi/setting_up_environment.md:42]:
```
3.  Should Lmod be unavailable and `$MODULEPATH` not used, you can initialise EESSI via an Lmod module by directly sourcing the Lmod initialisation script (this script automatically loads the EESSI module):
```

docs/using_eessi/setting_up_environment.md
boegel: or use `$SHELL` instead of harcoding `bash`?

docs/using_eessi/setting_up_environment.md
boegel: + how to check (`echo $SHELL`)

boegel (changes requested):

docs/using_eessi/setting_up_environment.md
ocaisa: I checked yesterday and `$SHELL` is not reliable for the supported shells, `$0` is more reliable but not very visually helpful

ocaisa: (or rather the even worse `$(basename $0)`)

docs/using_eessi/setting_up_environment.md
ocaisa: Done in https://github.com/EESSI/docs/pull/317/commits/ca529f3c0915e116b7f7e7b8556dfe9c24444002

PR #326, opened 2024-10-14
Minor changes to bot documentation

Neves-P:
This PR does a minor update to the bot documentation. It doesn't add new sections, but simply updates/corrects older information.

In particular: the examples no longer show the pilot repository but instead target `software.eessi.io` and the mention to the test suite now states that tests run depending on the built software (instead of stating that testing will be added to the workflow at a later date).


laraPPr (approved):
lgtm

PR #338, opened 2024-11-11
Remove instruction to fetch repo config from EESSI/filesystem layer 

casparvl:
As we are now part of the default repos in CVMFS, this should no longer be needed (I think). Currently double checking in the #docs channel.


casparvl:
@bedroge commented:

On a "clean" system, at least. The cvmfs package has a cvmfs-config dependency, and by default (if no config package is installed yet) it's satisfied by installing cvmfs-config-default. You could override this by installing cvmfs-config-none (which is an empty config package) or some other config package first/at the same time, and in that case you do need to add the EESSI configuration.

@boegel commented: 

That said, there are reasons to not rely on the cvmfs-config repo (firewall crap, etc.).
I think we should keep recommended installing a local CernVM-FS configuration for EESSI when possible

So, closing this PR, as we'll keep this in the docs intentionally.

PR #345, opened 2024-11-15
Documentation improvements for setting up a private Stratum 1 and configuring clients to use them

bedroge:
This (hopefully :smile:  ) addresses the feedback from @alainvanhoof. Also extended some related parts on that page a bit, e.g. by adding an example of the `hosts` file.


alainvanhoof:
Yes, this adds a lot more insights on what needs to be done to run a local stratum 1 successfully.

PR #368, opened 2025-01-02
Including Norwegian Ai cloud

Sabryr:
(new pull request to replace #365
 have included Norwegian Ai cloud as a site. We have an interface (https://orchestrator.naic.no/) where users can request VMs on local OpenStack based cloud, Google or on Azure. We use Terraform for the provisioning and then use Cloud init to preconfigure EESSI. So the users get a VM with EESSI redy to use. This helps us keep a uniform setup and increase reproducibility. i.e. if a workflow developed on one VM , it will work on another


trz42 (changes requested):
Thanks @Sabryr 

Suggested a little tweaking of the wording.

trz42 (approved):
Thanks! lgtm

PR #369, opened 2025-01-08
Adding an MPI Hello World example in documentation on how to run on top of EESSI without EB

hvelab:
To make an addition to https://github.com/EESSI/docs/pull/175

### DIFF
--- a/docs/using_eessi/building_on_eessi.md
+++ b/docs/using_eessi/building_on_eessi.md
@@ -191,7 +191,7 @@ Now, set the `LD_RUN_PATH` environment variable for all the libraries to point t
 {EESSI 2023.06} $ mpicc -o HelloWorld mpi.c
 ```
 
-This is the moment to check if the compiler picked all the libraries from the compatibility layer.
+This is the moment to check if the compiler picked all the libraries from the software and compatibility layer, not the host.
 
 Look at the difference on the library solving when using the compatibility layer ldd from the host one:
 
--- a/docs/using_eessi/building_on_eessi.md
+++ b/docs/using_eessi/building_on_eessi.md
@@ -181,7 +181,7 @@ $ source /cvmfs/software.eessi.io/versions/2023.06/init/bash
 {EESSI 2023.06} $ m̀l gompi/2023b
 ```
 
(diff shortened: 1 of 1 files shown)


docs/using_eessi/building_on_eessi.md
```diff
@@ -143,13 +143,98 @@ shell before building. To do this:
...
+
+```
+# Starting the environment
+$ source /cvmfs/software.eessi.io/versions/2023.06/init/bash
+
+# Loading the toolchain
+{EESSI 2023.06} $ m̀l gompi/2023b
+```
+
+Now, set the `LD_RUN_PATH` environment variable for all the libraries to point to the runtime libraries, then compile the code. From the previous text, we are assuming that you are using GCC as the compiler, but please change the `$EBROOTGCCCORE variable accordingly in case you are using any other kind of compiler that you want to prevent picking from the host.
```
hvelab, line 184: Not sure if to state the last part, I mean, Ideally everybody should use GCC to compile with EESSI...

docs/using_eessi/building_on_eessi.md
```diff
@@ -143,13 +143,98 @@ shell before building. To do this:
...
   ```sh
-      export LD_RUN_PATH=$LIBRARY_PATH:$EBROOTGCCcore/lib64
+      export LD_RUN_PATH=$LIBRARY_PATH:$EBROOTGCCCORE/lib64
   ```
 * Compile and make sure the library resolution points to the EESSI stack. For this, `ldd` from compatibility layer and **not** `/usr/bin/ldd` should be used
   when checking the binary.
 
 * Run! 
 
+To exemplify this, take the classic MPI Hello World example code:
```
hvelab, line 153: Thought to put it apart from the step-by-step explanation as it might be too verbose to put inside bullet points, any suggestion on how to do it better is appreciated

ocaisa (changes requested):

docs/using_eessi/building_on_eessi.md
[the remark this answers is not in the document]
hvelab: yup, I was not sure if adding this part as stated here https://github.com/EESSI/docs/pull/369/#discussion_r1907033370, 
closing the comment and commiting the change

ocaisa (changes requested):
A couple of final suggestions

ocaisa (approved):
LGTM

PR #372, opened 2025-01-15

EESSIbot:


boegel:
It's a bit too early to already include `sapphire_rapids`, we should exclude it for now => https://github.com/EESSI/docs/pull/373

PR #306, opened 2024-10-04
add zen4 as supported CPU target

boegel:

PR #374, opened 2025-01-24
add zen4 to supported CPU targets

bedroge:


bedroge:
Ah, duplicate of #306 .

PR #385, opened 2025-02-19
Add `ld+json` metadata to software detail pages

### LINKED ISSUE #383: We can make EESSI more FAIR by adding schema.org metadata to our `<head>`
Specifically I would like to target the available software and inject  https://schema.org/SoftwareApplication metadata into the `<head>` for each application under https://www.eessi.io/docs/available_software/detail .

Since these pages are dynamically generated, it should be possible to do this automatically. We can add the creation of json metadata to our script and then inject that into the yaml front matter. For this we would need to create a MkDocs plugin that extracts the header data and injects it, something like
```python
import json
from mkdocs.plugins import BasePlugin
from jinja2 import Markup

class InjectLDJsonPlugin(BasePlugin):
    def on_page_context(self, context, page, config, nav):
        """
        Inject per-page JSON-LD into the HTML head if available in front matter.
        """
        json_ld_data = page.meta.get("json_ld")  # Extract from front matter
        if json_ld_data:
            json_ld_script = f"""
            <script type="application/ld+json">
            {json.dumps(json_ld_data, indent=2)}
            </script>
            """
            if 'extrahead' in context:
                context['extrahead'] += Markup(json_ld_script)
            else:
                context['extrahead'] = Markup(json_ld_script)
        return context
``` 
and then
```yaml
plugins:
  - search
  - inject_ld_json
 ```

### PR DESCRIPTION
**Pull request by ocaisa**
Fixes #383

### CODE REVIEW

File: mkdocs-ldjson-plugin/setup.py [hunk 1]
```diff
@@ -0,0 +1,20 @@
+from setuptools import setup, find_packages
+
+# Read requirements from requirements.txt
+def read_requirements():
+    with open("requirements.txt", encoding="utf-8") as f:
+        return f.read().splitlines()
+
+setup(
+    name="inject_ld_json",
```
boegel, line 9: @ocaisa Maybe it makes sense to throw this in a dedicated repo, and publish it on PyPI, so others can also use this plugin, but good enough to park it in our `EESSI/docs` repo for now

boegel (approved):

PR #390, opened 2025-02-21

EESSIbot:


bedroge:
I had opened a PR to prevent this (https://github.com/EESSI/docs/pull/389), but apparently used `sapphire_rapids` again instead of `sapphirerapids`. Fixed in https://github.com/EESSI/docs/pull/391. This can be closed, we'll add `sapphirerapids` once the stack is complete.

PR #394, opened 2025-02-28
Fix font-size

laraPPr:
Strip all the unnecessary css from https://cdnjs.cloudflare.com/ajax/libs/fomantic-ui/2.9.2/semantic.min.css. So that only the table is affected by these style elements and not the font sizes on all the pages.


boegel (approved):
tested locally, lgtm

PR #395, opened 2025-03-01

EESSIbot:


ocaisa (approved):
We finally get to see the json-ld data!

ocaisa:
And it works: https://search.google.com/test/rich-results/result?id=KE5D95wqXROwc64wqTWQPw

PR #415, opened 2025-03-22
exclude aarch64/nvidia/grace from software overview updates

trz42:
Exclude `aarch64/nvidia/grace` from software overview updates while we build up the stack.


logo/EESSI_logo_horizontal_white.jpg
bedroge, line 1: Was it your intention to also change the logo?

trz42: Nope

trz42: fixed in https://github.com/EESSI/docs/pull/415/commits/4ea43ee4862795822be028d18a16686ffaefbf45

thanks for checking 🤗

bedroge (approved):

scripts/available_software/available_software.py
```diff
@@ -28,7 +28,7 @@
 EESSI_TOPDIR = "/cvmfs/software.eessi.io/versions/2023.06"
 
 # some CPU targets are excluded for now, because software layer is too incomplete currently
-EXCLUDE_CPU_TARGETS = ['aarch64/a64fx']
+EXCLUDE_CPU_TARGETS = ['aarch64/a64fx', 'aarch64/nvidia/grace']
```
bedroge, line 31: @trz42 Apparently this didn't work, see #416. We may need to do some work for the extra subdir, but for now we can maybe just exclude `aarch64/nvidia`?

PR #428, opened 2025-04-04
training section with EESSI webinar series may/june 2025

trz42:
New training top-level section and website for EESSI webinar series in May/June 2025.


ocaisa (approved):
LGTM

PR #439, opened 2025-04-25

EESSIbot:


bedroge:
Another failed attempt to include `aarch64/nvidia/grace`. Though the json now does include it, it also still included `aarch64/nvidia`, which it should ignore/skip.

Issue #402, opened 2025-03-03
Automatically generate the columns for the software overview page


bedroge:
I initially forgot to update https://github.com/EESSI/docs/blob/main/docs/available_software/overview.md, and then a merge of the automatically generated overview pages https://github.com/EESSI/docs/pull/399) basically broke the entire page: no software was listed anymore, because the columns didn't match. It would be nice if the columns on the main page can be auto-generated or something to prevent that from happening.

PR #441, opened 2025-04-25

EESSIbot:


bedroge:
Though this worked and included `grace`, there are issues with the ordering, see #402. Due to this we removed Grace again from the overview in #443, so let's close this PR.

PR #442, opened 2025-04-25
Temporary workaround that hardcodes the list of targets in the Python code

bedroge:
This makes sure that we get the columns ordered in the most clear way (generic first, etc), which aligns with the headers that are defined in the markdown file. It would be nice if the javascript code would match the table headers with the json data, see https://github.com/EESSI/docs/issues/402.

### DIFF
--- a/docs/available_software/overview.md
+++ b/docs/available_software/overview.md
@@ -11,7 +11,8 @@ This table gives an overview of all the available software in EESSI per specific
             <th colspan="7">x86_64</th>
         </tr>
         </tr>
-            <th colspan="4"></th>
+            <th colspan="3"></th>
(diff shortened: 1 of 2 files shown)


bedroge:
Screenshot of a locally rendered page:

image

boegel:
Is this still relevant?

If so, it needs work (to resolve the conflict)

cc @laraPPr

laraPPr:
This one should just be closed because we don't need the to hardcode the targets since where now using this compare function

https://github.com/EESSI/docs/blob/333ffac82444b1604be2a15910bc4b24627c7970/scripts/available_software/available_software.py#L224C5-L224C26

PR #443, opened 2025-04-25
exclude Grace again from software overview

bedroge:
The page is currently broken, so let's revert the changes from #436 .


laraPPr (approved):
lgtm

PR #445, opened 2025-04-30
Test suite apps and change in the minimum Python version 

satishskamath:
- Added all the apps currently available via the test-suite.
- Changed minimum Python version from 3.6 to 3.7 for the test-suite Reframe.

### DIFF
--- a/docs/test-suite/installation-configuration.md
+++ b/docs/test-suite/installation-configuration.md
@@ -14,7 +14,13 @@ The EESSI test suite requires
 
 ??? note "(If your system Python version is <= 3.6, click here for some tips)
 
-    * You can upgrade the python version in a virtual environment. [Link](https://docs.python.org/3/library/venv.html)
+    * You can upgrade and manage python versions using `pyenv`. [Link](https://github.com/pyenv/pyenv?tab=readme-ov-file#installation)
+        * For installation follow the documentation in the repository using the link above.
+        * Then follow the steps to install a new Python version and further change local paths to the new Python:
+
+                pyenv install <python version number>
+                pyenv local <python version number>
+
     * You can also install a more recent version of Python on top of the GCC/GCCcore compiler.
 
 #### Installing Reframe
--- a/docs/test-suite/installation-configuration.md
+++ b/docs/test-suite/installation-configuration.md
@@ -8,9 +8,14 @@ This page covers the requirements, installation and configuration of the [EESSI 
 
 The EESSI test suite requires 
 
-* Python >= 3.6 
+* Python >= 3.7
 * [ReFrame](https://reframe-hpc.readthedocs.io) v4.3.3 (or newer)
 * [ReFrame test library (`hpctestlib`)](https://reframe-hpc.readthedocs.io/en/stable/hpctestlib.html)
+
+??? note "(If your system Python version is <= 3.6, click here for some tips)
+
+    * You can upgrade the python version in a virtual environment. [Link](https://docs.python.org/3/library/venv.html)
+    * You can also install a more recent version of Python on top of the GCC/GCCcore compiler.
 
 #### Installing Reframe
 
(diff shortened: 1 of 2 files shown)


docs/test-suite/installation-configuration.md
```diff
@@ -8,10 +8,15 @@ This page covers the requirements, installation and configuration of the [EESSI
...
 The EESSI test suite requires 
 
-* Python >= 3.6 
+* Python >= 3.7
 * [ReFrame](https://reframe-hpc.readthedocs.io) v4.3.3 (or newer)
 * [ReFrame test library (`hpctestlib`)](https://reframe-hpc.readthedocs.io/en/stable/hpctestlib.html)
 
+??? note "(If your system Python version is <= 3.6, click here for some tips)
+
+    * You can upgrade the python version in a virtual environment. [Link](https://docs.python.org/3/library/venv.html)
```
casparvl, line 17: Wait, what? You cannot install a newer python in a virtualenv, right?

You can (according to my AI friend) do this with `pyenv` https://github.com/pyenv/pyenv. I think that's also the tool that Sam referred to. As far as I understand, `pyenv` installs into your homedir by default (i.e. perfect for userspace installs) and allows you to install new python versions with `pyenv install <version>; pyenv global <version>  # Sets tje default python version to use for your user; pyenv --version`. Note: I haven't tried myself, could be good to give it ago before you put it in the docs to be sure this works. I'd probably just say: "Install pyenv by following their documentation https://github.com/pyenv/pyenv", and then give the explicit instruction on how to install a different python version.

satishskamath: Yeah, true. I read it wrong in the stackoverflow answer and in the documentation as well. :D The documentation states that one can upgrade the packages in the venv to a newer version as long as a newer python version is in place.
Indeed what Sam was talking was about `pyenv`.
This is only possible with a local package manager such as `conda`. Not sure if we want to recommend `conda` here.

satishskamath: By "This" I meant upgrading the python version. It is possible using
1. `pyenv` indeed does not require existing python versions, it is just a python version manager. I just tested it and it works.
2. `conda` is a second alternative but I was not sure whether we want to pitch conda. :)

satishskamath:
@smoors and @casparvl  both your suggestions are incorporated.

docs/test-suite/installation-configuration.md
```diff
@@ -8,10 +8,22 @@ This page covers the requirements, installation and configuration of the [EESSI
...
+??? note "(If your system Python version is lower than the minimum required version, click here for some tips)
+
+    * You can upgrade and manage python versions using `pyenv`. [Link](https://github.com/pyenv/pyenv?tab=readme-ov-file#installation)
+        * For installation follow the documentation in the repository using the link above.
+        * Then follow the steps to install a new Python version and further change local paths to the new Python:
+
+                pyenv install <python version number>
+                pyenv local <python version number>
+
+    * You can install a more recent version of Python on top of the GCC/GCCcore compiler.
```
smoors, line 24: don't forget to set `RFM_PURGE_ENVIRONMENT=1` is you use Python from a module. the ReFrame easyconfigs already do that for you.

satishskamath: I do not see that env variable [here](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/r/ReFrame/ReFrame-4.6.2.eb) nor in my module env of ReFrame which was installed using EB.

smoors: that's because your easyconfig uses the system python.

see here for an example that uses another python: https://github.com/easybuilders/easybuild-easyconfigs/blob/0835a1a062a1bbc8777a5a4164dc2e3fbf6d907b/easybuild/easyconfigs/r/ReFrame/ReFrame-4.6.3-GCCcore-13.3.0.eb#L71

satishskamath: aah ok. Will change this.

casparvl (approved):

boegel:
@satishskamath There's a bullet list on top of the page, that should be updated too to link to the subsections?

see https://www.eessi.io/docs/test-suite/available-tests/

satishskamath:
@boegel yeah, just took care of it with https://github.com/EESSI/docs/pull/465 . Thanks. :)

PR #447, opened 2025-05-02
Order the targets in available_software.py

laraPPr:
Also needed to fix the tests to take the ordering into account


docs/available_software/overview.md
```diff
@@ -7,26 +7,26 @@ This table gives an overview of all the available software in EESSI per specific
...
-            <th colspan="1">generic</th>
-            <th colspan="1">neoverse_n1</th>
-            <th colspan="1">neoverse_v1</th>
             <th colspan="1">generic</th>
             <th colspan="1">zen2</th>
             <th colspan="1">zen3</th>
             <th colspan="1">zen4</th>
             <th colspan="1">haswell</th>
             <th colspan="1">skylake_avx512</th>
             <th colspan="1">sapphirerapids</th>
```
bedroge, line 26: The script now sorts these alphabetically, right? Then these would need to be reversed:
```suggestion
            <th colspan="1">sapphirerapids</th>
            <th colspan="1">skylake_avx512</th>
```

laraPPr: I've now removed all the names since javascript overrides it. I'm now writing a test to check that there are the right number of colums.

laraPPr:
Tested new workflow here. I hope the workflow will be triggered for the update available scripts pr once this pr gets merged.

laraPPr:
And the tests of the available_software.py also do not seem to be running but I ran them locally.

laraPPr:
This workflow is also not running even though I think it should be triggered [.github](https://github.com/EESSI/docs/tree/main/.github)/[workflows](https://github.com/EESSI/docs/tree/main/.github/workflows)
/script_module_list.yml

laraPPr:
The test for the changed available_software.py script are ran see https://github.com/EESSI/docs/actions/runs/14799376763

scripts/available_software/available_software.py
```diff
@@ -233,7 +233,33 @@ def modules_eessi() -> dict:
...
+            else:
+                targets_x86_64.append(target)
+    ordered_targets = []
+    ordered_targets.extend(generic_target_x86_64)
+    ordered_targets.extend(np.sort(targets_x86_64))
+    ordered_targets.extend(generic_target_aarch64)
+    ordered_targets.extend(np.sort(targets_aarch64))
+
+    targets = [t for t in ordered_targets if not any(t.endswith(x) for x in EXCLUDE_CPU_TARGETS)]
+
```
bedroge, line 262: I think this does more or less what we want for now, but I was wondering if we could make this a bit more dynamic (e.g. because we will probably add RISC-V in the future). I

It looks like we could use a single `find` command to find all the targets:
```
find ${EESSI_TOPDIR}/software/linux -maxdepth 4 -name modules -type d -exec dirname {} \;
```
(this is also used here: https://github.com/EESSI/filesystem-layer/blob/main/scripts/update_lmod_caches.sh#L37)

Once you have them all in a list, we could perhaps use a custom compare function for the sorting, and make sure that we always list `generic` first, followed by the ones with a vendor dir (`amd/intel/nvidia`), followed by the other ones. I quickly put this together and this seems to do the trick: 
```
# got this list with the find command
targets = [
'/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1',
'/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_v1',
'/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic',
'/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/a64fx',
'/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/nvidia/grace',
'/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2',
'/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen3',
'/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen4',
'/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/generic',
'/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/haswell',
'/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/skylake_avx512',
'/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/sapphirerapids',
'/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/cascadelake',
'/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/icelake',
]


def eessi_target_compare(a, b):
    if a == b:
        return 0
    
    a_split = a.rsplit('/')
    b_split = b.rsplit('/')
  
    # We first compare the main architecture (aarch64, x86_64), which is the 7th field
    if a_split[7] == b_split[7]:
        # Check if one item is for generic builds (last field), these should always be listed first
        if a_split[-1] == 'generic':
            return -1
        if b_split[-1] == 'generic':
            return 1
        # If the number of fields are not equal, one has an extra vendor subdirectory (e.g. amd, intel, nvidia).
        # These should always come after the ones without this extra level.
        if len(a_split) != len(b_split):
            return 1 if len(a_split) > len(b_split) else -1

      # In all other cases we just do an alphabetical sort of the strings.
      return 1 if a > b else -1

from functools import cmp_to_key
eessi_target_compare_key = cmp_to_key(eessi_target_compare)

sorted(targets, key=eessi_target_compare_key)
# or targets.sort(key=eessi_target_compare_key)
```

laraPPr: I've implemented the suggested funtion

scripts/available_software/available_software.py
```diff
@@ -220,6 +221,33 @@ def targets_eessi() -> np.ndarray:
     return targets
 
 
+def eessi_target_compare(a, b):
+    """
+    A comparison function to compare the EESSI targets and order them
+    @return: 0, 1, -1
+    """
```
Neves-P, line 228:
```suggestion
    """
    A comparison function to compare the EESSI targets and order them.
    
    First the main architecture is ordered alphabetically, then within them
    the CPU targets are again ordered alphabetically, except for the 
    generic target, which always comes first. Targets that include an extra
    vendor subdir always after those without a vendor subdir.
    @return: 0, 1, -1
    """
```

.github/workflows/test_overview_available_software.yml
Neves-P, line 1: This workflow didn't run (as expected, given run the conditions), but it did run in @laraPPr's branch and works well: https://github.com/laraPPr/EESSI_docs/actions/runs/14799375963, so it's safe to assume it will run well when the conditions are met in the target branch.

Neves-P (changes requested):
@laraPPr, looks good! I just made a small suggestion on the docs of a function in the docstring. Feel free to reject if you prefer the info to stay as is (since you already have code comments)  :+1:

laraPPr:
The linter is failing on yo docstring

scripts/available_software/available_software.py
```diff
@@ -220,6 +221,38 @@ def targets_eessi() -> np.ndarray:
     return targets
 
 
+def eessi_target_compare(a, b):
+    """
+    A comparison function to compare the EESSI targets and order them.
+
```
laraPPr, line 227:
```suggestion

```

scripts/available_software/available_software.py
```diff
@@ -220,6 +221,38 @@ def targets_eessi() -> np.ndarray:
     return targets
 
 
+def eessi_target_compare(a, b):
+    """
+    A comparison function to compare the EESSI targets and order them.
+    
+    First the main architecture is ordered alphabetically, then within them
+    the CPU targets are again ordered alphabetically, except for the
```
laraPPr, line 229:
```suggestion
    the CPU targets are again ordered alphabetically, except for the
```

Neves-P:
> The linter is failing on yo docstring

Sorry about that! Pushed your suggestions now, hopefully it passes now :+1:

PR #452, opened 2025-05-05
Some improvements to tutorial cvmfs section

HereThereBeDragons:
Please DO NOT merge before yo uhave taken a decision on:
`<span style="color: red;">data vs files? would keep it consistent</span>`
in `docs/tutorial/cvmfs/what-is-cvmfs.md`

### DIFF
--- a/docs/tutorial/cvmfs/what-is-cvmfs.md
+++ b/docs/tutorial/cvmfs/what-is-cvmfs.md
@@ -35,6 +35,8 @@ support this, including:
 * [compression of data](#features-compression);
 * [verification of data integrity](#features-data-integrity);
 
+<span style="color: red;">data vs files? would keep it consistent</span>
+
 CernVM-FS has been proven to scale to *billions* of files and *tens of thousands* of [clients](../appendix/terminology.md#client).
 
 It was originally developed at [CERN](https://home.cern/) to let High Energy Physics (HEP) collaborations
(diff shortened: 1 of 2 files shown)


docs/tutorial/cvmfs/what-is-cvmfs.md
```diff
@@ -35,6 +35,8 @@ support this, including:
 * [compression of data](#features-compression);
 * [verification of data integrity](#features-data-integrity);
 
+<span style="color: red;">data vs files? would keep it consistent</span>
```
boegel, line 38: @HereThereBeDragons Well, I guess it depends on the context...

Server-side, "data" makes more sense, since CernVM-FS stores the software installations in a hash tree.
Technically they're just files, but there's "useless" files when accessed directly.

"Files" makes more sense when you're referring to what a CernVM-FS client exposes to end users.

docs/tutorial/cvmfs/what-is-cvmfs.md
```diff
@@ -115,6 +117,8 @@ This can result in a significant reduction in storage capacity that is required
 especially when identical files are spread out across the repository,
 as often happens with particular files like example data files across multiple versions of the same software.
 
+For the CVMFS client, data de-duplication will also happen across repositories when the local client cached is shared across all repositories (the default).
```
boegel, line 120:
```suggestion
For the CernVM-FS client, data de-duplication will also happen across repositories when the local client cached is shared across all repositories (the default).

Comment (boegel) [inline docs/tutorial/cvmfs/what-is-cvmfs.md:120]:
```suggestion
For the CernVM-FS client, data de-duplication will also happen across repositories when the local client cached is shared across all repositories (the default).

**Review by boegel — CHANGES_REQUESTED** [review]

File: docs/tutorial/cvmfs/what-is-cvmfs.md (hunk 1 above)
Comment (boegel) [inline docs/tutorial/cvmfs/what-is-cvmfs.md:38]: I've opened an issue to follow-up on it (see #455), so we can go ahead and merge, after discussing it with @HereThereBeDragons

```suggestion
```

boegel (changes requested):

PR #459, opened 2025-05-08
some small fixes to the client

HereThereBeDragons:
please DO NOT MERGE until you have taken a decision on and removed comments marked with `!!! danger`

### DIFF
--- a/docs/tutorial/access/client.md
+++ b/docs/tutorial/access/client.md
@@ -82,7 +82,7 @@ which should contain something like:
 
 ``` { .ini .copy }
 CVMFS_CLIENT_PROFILE="single" # a single node setup, not a cluster
-CVMFS_QUOTA_LIMIT=10000
+CVMFS_QUOTA_LIMIT=10000       # 10 GB of local client cache, shared
 ```
 
 More information on the structure of `/etc/cvmfs` and supported configuration settings is available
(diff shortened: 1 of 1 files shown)


docs/tutorial/access/client.md
```diff
@@ -351,6 +358,8 @@ To start using EESSI just [source the initialisation script](../eessi/using-eess
 source /cvmfs/software.eessi.io/versions/2023.06/init/bash
 ```
 
+!!! danger "update link to latest version? maybe add to eessi a symlink /cvmfs/software.eessi.io/versions/latest/init/bash"
```
boegel, line 361:
```suggestion
```

For now this is still the only version we had. We actually used to have a `latest` symlink, but we stopped doing that because figuring out when to update the target of that is ... challenging.

boegel (changes requested):

docs/tutorial/access/client.md
vvolkl: Sounds reasonable. And for the tutorial even an old version would work.

docs/tutorial/access/client.md
```diff
@@ -226,6 +231,8 @@ update the `autofs` configuration.
 
 ### Recommendations for `autofs` {: #autofs }
 
+!!! danger "is this still recommended?"
```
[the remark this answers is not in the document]
vvolkl: No, the linked issue does no longer apply, I think. There may be circumstances with bind-mounts into containers, let me double-check that

PR #469, opened 2025-05-23
Add Intel Cascade Lake and Ice Lake

bedroge:
Also needs https://github.com/EESSI/software-layer/pull/1084 to be merged around the same time to actually detect these CPU targets.


bedroge:
Note that I put Cooper Lake at the same level as Cascade Lake. It's actually a bit newer (released at the same time as Ice Lake?), but Ice Lake seems to support more instructions:
https://en.wikichip.org/wiki/intel/microarchitectures/cooper_lake#New_instructions
https://en.wikichip.org/wiki/intel/microarchitectures/ice_lake_(server)#New_instructions

boegel (approved):
lgtm

PR #479, opened 2025-06-05
Update policy away from `from-pr` and use `from-commit` instead

Neves-P:
This change and minor version update of the contribution policy to reflect that we require `from-commit` instead of `from-pr` on new contributions. I assumed a change in the text implies an increase in the policy's version, if not, let me know and I can revert those changes.

Thanks, @pescobar for spotting it!


boegel (approved):
lgtm, makes perfect sense, thanks @Neves-P (and @pescobar for flagging this)

PR #490, opened 2025-06-24
ROCm blog post

Zeldhoron:


docs/blog/posts/2025/06/eessi-rocm.md
```diff
@@ -0,0 +1,80 @@
...
+* [Jan André Reuter](https://github.com/Thyre)'s custom EasyBuild support for ROCm available [here](https://github.com/Thyre/easybuild-custom/tree/support-passing-amdgcn/easybuild/easyconfigs/r)
+
+A crucial contribution came from [Davide Grassano](https://github.com/Crivella) who successfully got ROCm-LLVM to build with EasyBuild.
+This involved significant technical work, including contributions back to the EasyBuild project through pull requests [#3706](https://github.com/easybuilders/easybuild-easyblocks/pull/3706) and [#3781](https://github.com/easybuilders/easybuild-easyblocks/pull/3781).
+Getting ROCm-LLVM working was essential because it serves as the compiler foundation for the entire ROCm stack.
+
+## Current Progress: Building the Core Stack
+
+We've made substantial progress in building ROCm components for version 6.4.0 within the EESSI build environment.
+The following core components now build successfully with EasyBuild in our EESSI build container: ROCm-LLVM, rocminfo, rocm-cmake, HIP, amdsmi, roctracer.
```
Thyre, line 51:
```suggestion
The following core components now build successfully with EasyBuild in our EESSI build container: ROCm-LLVM, rocminfo, rocm-cmake, HIP, amdsmi, ROCTracer.
```

AMD has like three or four different variants for the spelling of ROCTracer in their own repository, but the documentation uses ROCTracer. We should probably stick with that.

docs/blog/posts/2025/06/eessi-rocm.md
```diff
@@ -0,0 +1,80 @@
...
+
+Currently, we're conducting our testing and building on a virtual machine environment without physical AMD GPU hardware.
+This approach allows us to validate the build process and catch configuration issues, though it means some sanity checks that require actual GPU hardware are necessarily skipped during the build process.
```
Thyre, line 54: Just as a note, something I noticed while going through the ROCm stack myself:
Most, if not all, of the ROCm stack I've handled so far doesn't actually check the hardware to build. Some packages have an option for this, but this is disabled by default.

Once we try to enable testing (e.g. `ctest`), actual hardware is required for the tests to work. Depending on the hardware, there's also a high likelihood that some tests fail. Don't know yet how we want to handle this.

docs/blog/posts/2025/06/eessi-rocm.md
```diff
@@ -0,0 +1,80 @@
...
+We haven't yet performed runtime validation with actual AMD GPU hardware, which means we can't fully verify that everything works as expected in real-world scenarios.
+Additionally, some sanity checks that require GPU presence are skipped during the build process, so we're building with partial validation.
+
+## Next Steps: Validating the Integration
+
+Looking ahead, we've identified clear priorities for completing our ROCm integration work.
+
+Our high-priority next steps focus on validation and completeness.
+We need to add proper support for AMD GPU drivers and hardware detection, ensuring our built components can actually communicate with AMD GPUs.
+We're also working to finish adding support for remaining core components and AMD's validation suite and examples that demonstrate real-world functionality.
```
Thyre, line 69: Maybe link to both the ROCmValidationSuite repository and ROCm-examples?
Worth noting that the full ROCm-examples repository basically requires all the components. ROCmValiationSuite is a somewhat easy target to reach, and I've verified that my EasyConfigs work on both `gfx1201` (on Arch Linux with a system ROCm present as well) and `gfx90a` (in the [EasyBuild Rocky Linux 9.5](https://github.com/easybuilders/easybuild-containers/pkgs/container/rockylinux-9.5) container without ROCm) at least.

docs/blog/posts/2025/06/eessi-rocm.md
```diff
@@ -0,0 +1,80 @@
...
+
+A crucial milestone will be running the validation suite and examples on actual AMD GPU hardware to verify that everything works correctly.
+
+Our lower-priority items, while still important, focus on expanding capabilities and improving documentation.
+We plan to add support for additional libraries and frameworks beyond the core stack, integrate popular scientific applications that leverage ROCm, and update our ROCm overview document with practical insights gained from our building and testing experience.
+We're also exploring the possibility of contributing our overview back to AMD's official documentation, which could benefit the broader community.
```
Thyre, line 75: One crucial part should probably also providing feedback to improve building ROCm. This could be quite important for AMD, as they're working on a new "build-from-scratch" stack called [TheRock](https://github.com/ROCm/TheRock).

docs/blog/posts/2025/06/eessi-rocm.md
Thyre: Since the ROCm examples are similarly structured to the CUDA examples, we unfortunately cannot disable certain parts easily. I agree though that we can write up some documentation to explain how users can run the examples.

Let's see first if we can build all the required modules though.

docs/blog/posts/2025/06/eessi-rocm.md
Thyre: One of the options would be to handle it similarly to LLVM. Maybe have some threshold we define in EasyConfigs for a percentage / flat amount of tests we expect to see failing.
I'll start enabling tests for my EasyConfigs soonish (once I find some time for it). Then, we may get a better sense of if (and how many) tests are failing.

For officially supported architectures, I expect this number to be quite small. If one tries to build for example for a not officially supported arch like `gfx1152`, I would expect to see quite some failures.
In EasyBuild, one could simply specify `--ignore-test-failure`. EESSI would probably not include support for these architectures.

boegel (approved):

PR #499, opened 2025-07-01
Add notes for CVMFS repos that might be used as URLs

ocaisa:


ocaisa:
The redirects for software.eessi.io dev.eessi.io and riscv.eessi.io should all be working now.

PR #500, opened 2025-07-01
Remove references to pilot repository

bedroge:


bedroge:
> There is still a mention of the pilot in docs/tutorial/eessi/high-level-design/#filesystem_layer. I think this should also be moved, or maybe we want to keep some historical record of the development? In that case we should rephrase this to make it clear that the pilot is no longer available.

Thanks, good catch! I had ignored that subdir, but you're right that there's no point in leaving that note there, so I've removed it.

Neves-P (approved):
Looks good to me!

PR #501, opened 2025-07-01
Update adding new software instructions after splitting software-layer and software-layer-scripts

Neves-P:
This PR changes and adds information to reflect that the build scripts are now in EESSI/software-layer-scripts and that new PRs should target the main branch of EESSI/software-layer.

It also fixes a few small inconsistencies in the text.

Rendered docs on my fork: https://neves-p.github.io/docs/adding_software/building_software/ and https://neves-p.github.io/docs/adding_software/opening_pr/


casparvl:
Out of scope for this PR (don't hold it back for this!), but: I think we should also document what to do if people want to contribute to the `software-layer-scripts`.

docs/adding_software/opening_pr.md
```diff
@@ -41,24 +41,34 @@ git remote add koala git@github.com:koala/software-layer.git
 ```
 
 3) Check out the branch that corresponds to the version of EESSI repository you want to add software to,
```
casparvl, line 43: This whole sentence should be rephrased, probably just to 'Make sure you're on the main branch:' (I think that's useful to keep, even if that's also the default branch - if people already cloned the repo earlier, it might help them remember that _at least_ they should switch back to the main branch).

docs/adding_software/opening_pr.md
```diff
@@ -91,20 +101,19 @@ git push koala example_branch
 6) Go to the [GitHub web interface](https://github.com/EESSI/software-layer) to open your pull request,
    or use the helpful link that should show up in the output of the `git push` command.
 
-   **Make sure you target the correct branch**: the one that corresponds to the version of EESSI you want to add
-   software to (like `2023.06-software.eessi.io`).
+   **Make sure you target the `main` branch.**
 
    If all goes well, one or more bots :robot: should almost instantly create a comment in your pull request
```
casparvl, line 106: This is no longer true, since we've silenced all bots. So this should be removed, otherwise people may think something is wrong.

casparvl (changes requested):
> echo '  - example-1.2.3-GCC-12.3.0.eb' >> easystacks/software.eessi.io/2023.06/eessi-2023.06-eb-4.8.2-2023a.yml

Should also be updated. Moreover, I think that section (point 3) should more clearly describe the location of the easystack files - and the fact that those are (now) in a version specific subdir.

> git add easystacks/software.eessi.io/2023.06/eessi-2023.06-eb-4.8.2-2023a.yml

Also needs an update wrt the location of the easystack files.

docs/adding_software/opening_pr.md
Neves-P: Changed in 24b7cfe.

casparvl (approved):

PR #514, opened 2025-07-17
Add more details for EESSI CI/CD hackathon

ocaisa:


ocaisa:
Pushing this through so I can hit send on an associated email

PR #522, opened 2025-08-06
Update bot build instructions for new node_type_map update

casparvl:
I think this covers all of the behavior that has changed in https://github.com/EESSI/eessi-bot-software-layer/pull/312

### DIFF
--- a/docs/bot.md
+++ b/docs/bot.md
@@ -117,7 +117,7 @@ bot: build repo:eessi.io-2023.06-software for:arch=x86_64/amd/zen4 inst:azure
 
 #### Native builds
 
-If you want to allocate the same node type that you want to build for, you can ommit the `on:` argument.
+If you want to allocate the same node type that you want to build for, you can omit the `on:` argument.
 For example: `bot:build for:arch=x86_64/amd/zen4` is fully equivalent to `bot:build on:arch=x86_64/amd/zen4 for:arch=x86_64/amd/zen4`.
 
 #### Cross-compiling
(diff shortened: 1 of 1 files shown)


docs/bot.md
```diff
@@ -108,17 +111,43 @@ repositories, and CPU targets. Specify one build instruction per line.
...
+bot: build repo:eessi.io-2023.06-software for:arch=x86_64/amd/zen3 inst:aws
+bot: build repo:eessi.io-2023.06-software for:arch=x86_64/amd/zen4 inst:azure
 ```
 
-!!! note
-    The bot applies the filters with partial matching, which you can use to combine multiple build
-    instructions into a single one.
+#### Native builds
+
+If you want to allocate the same node type that you want to build for, you can ommit the `on:` argument.
```
casparvl, line 120:
```suggestion
If you want to allocate the same node type that you want to build for, you can omit the `on:` argument.
```

docs/bot.md
```diff
@@ -66,13 +66,18 @@ should be issued by posting a comment in the pull request (see also [here](addin
...
 ```
 
-!!! warning
-    Only use `bot: build` if you are confident that it is OK to do so.
+Where the `for_arch` could be e.g. `x64_64/amd/zen4`. This will trigger the bot to allocate a node of that type _and_
+build in the `/cvmfs/software.eessi.io/versions/<eessi-version>/software/linux/x86_64/amd/zen4` prefix.
 
-    Most likely, you want to supply one or more filters to avoid that the bot builds for all its configurations.
+!!! note
+    The `for:` (and `on:`, see below) argument to the bot where introduced in bot version 0.9.0. They replace the `architecture=... accelerator=...` syntax used in bot versions <= v0.8.0.
```
bedroge, line 76:
```suggestion
    The `for:` (and `on:`, see below) argument to the bot were introduced in bot version 0.9.0. They replace the `architecture=... accelerator=...` syntax used in bot versions <= v0.8.0.
```

PR #548, opened 2025-09-09
Add UniBasel (scicore) to list of systems using EESSI

pescobar:
I added this entry following the alphabetical order for countries, but Czech Republic breaks it and my OCD hurts :)

Is it ok if I move the Czech republic up so every country is alphabetically ordered? Or if you prefer I can also move the scicore entry to the bottom?


boegel:
@pescobar Makes sense to restore the order in the universe and fix alphabetical ordering per country...

PR #559, opened 2025-09-17
Explain how to override architecture detection

ocaisa:


timeu:
Maybe also worth mentioning `EESSI_OVERRIDE_GPU_CHECK` because reframe will try to load the CUDA module and fail if it is run on a node without a GPU.

PR #562, opened 2025-09-22
Add University of Wuppertal

stderr-enst:
I've added our university cluster to the list of sites offering EESSI.
As an ATLAS site that ran CVMFS anyway, it was very straight forward for us to offer EESSI as well. :)


Neves-P (approved):
Looks good to me, thanks a lot, @stderr-enst !

PR #585, opened 2025-10-10
Add documentation on EESSI versions

Neves-P:
This is a draft PR to kick start work on EESSI versions.

It is currently a draft because a lot of information must still be added, and we may consider changing the structure or organization of some of the existing and new documentation pages to reflect the existence of different versions.

Feel free to push commits to this branch to expand the docs, or to modify what is already added.

_Note: this PR also adds a link and fixes a formatting issue in the systems.md page_


docs/versions/versions.md
```diff
@@ -0,0 +1,20 @@
+# EESSI Versions
+
+EESSI is organized in different dated versions. Each EESSI version contains a version of the [compatibility layer](compatibility_layer.md) and its' own set of software applications. In addition, a particular EESSI version will target a different set of [common toolchains](https://docs.easybuild.io/common-toolchains/) to build software with. As an example, the first production EESSI version was released in June 2023 and is named `2023.06`. It contains the production release of the compatibility layer used to build software, and contains software build using the  `foss/2022b`, `foss/2023a`, and `foss/2023b` [software toolchains](https://docs.easybuild.io/common-toolchains/#common_toolchains_overview) as defined in EasyBuild.
```
casparvl: I don't think people will understand what 'the production release' means. Isn't an EESSI version _essentially_ related to a fixed version of Gentoo-prefix for the compat layer? I mean, we apply security updates, but we don't update the version of Gentoo-prefix _itself_, or do we?

If my understanding is correct, I'd state something like: "Its compatibility layer is based on Gentoo-Prefix version XYZ, and the software layer containers software build using the ... toolchains".

PR #701, opened 2026-03-16
List support for AMD Zen5 CPU target

Neves-P:
This PR just adds at line about AMD Zen 5 support in https://www.eessi.io/docs/software_layer/cpu_targets/


bedroge (approved):

bedroge:
Ah, apparently I had already (prematurely) added Zen5 support as part of https://github.com/EESSI/docs/pull/734. So I guess we can close this PR.

PR #731, opened 2026-04-04
add note on lmod cache when loading an EESSI module

smoors:
```
$ module unuse $MODULEPATH
$ module use /cvmfs/software.eessi.io/init/modules
$ module load EESSI/2025.06                                                                                                                                       
Lmod has detected the following error:  Unable to load module because of error when evaluating modulefile:
     /cvmfs/software.eessi.io/init/modules/EESSI/2025.06.lua: /usr/share/lmod/lmod/libexec/ModuleA.lua:676: Did not find mpath in mA

     Please check the modulefile and especially if there is a line number specified in the above message
If you don't understand the warning or error, contact the helpdesk at hpc@vub.be 
While processing the following module(s):
    Module fullname  Module Filename
    ---------------  ---------------
    EESSI/2025.06    /cvmfs/software.eessi.io/init/modules/EESSI/2025.06.lua
```
```
$ LMOD_IGNORE_CACHE=1 module load EESSI/2025.06
Module for EESSI/2025.06 loaded successfully
```
```
$ module --version                                                                                                                                                                                 

Modules based on Lua: Version 8.7.65 2025-08-05 10:24 -06:00
    by Robert McLay mclay@tacc.utexas.edu
```

### DIFF
--- a/docs/using_eessi/setting_up_environment.md
+++ b/docs/using_eessi/setting_up_environment.md
@@ -62,6 +62,11 @@ available to use. If you would like to see what environment variables the module
         module load EESSI/2023.06
         ```
 
+        !!! Note
+            If you encounter an Lmod error (`Did not find mpath in mA`), you
+            may need to disable the Lmod cache by setting LMOD_IGNORE_CACHE=1
+            in your environment before loading the module.
+
         :clap: Your environment is now set up, you are ready to start running software provided by EESSI!
 
     2.  If you are using an Lmod with a version older than 8.6 or any other module tool utilizing `MODULEPATH` (e.g., 


smoors:
i did some testing and digging.

it's an issue with Lmod itself, specifically in the file `ModuleA.lua`, starting from 8.7.65 and fixed in 9.0.5

removing the original cache didn't help, and adding a cache for the EESSI modules didn't help either.

ocaisa:
This is a tricky thing to fix or work around. One thing we could do is have the module print a load message for that version range, that's better than doing it in the docs because they can act on it straight away.

ocaisa:
Hmm, there may be a way to get the behaviour we would need with `haveDynamicMPATH()`, let me see if I can reproduce

ocaisa:
I can't reproduce this, I've tried a few different Lmod versions in the range.

```
ocaisa@~/Lmod((HEAD detached at 8.7.65))$ module purge
ocaisa@~/Lmod((HEAD detached at 8.7.65))$ rm ~/.cache/lmod/ -r
rm: cannot remove '/home/ocaisa/.cache/lmod/': No such file or directory
ocaisa@~/Lmod((HEAD detached at 8.7.65))$ module unuse $MODULEPATH
ocaisa@~/Lmod((HEAD detached at 8.7.65))$ module use /cvmfs/software.eessi.io/init/modules
ocaisa@~/Lmod((HEAD detached at 8.7.65))$ module load EESSI/2025.06
Module for EESSI/2025.06 loaded successfully
ocaisa@~/Lmod((HEAD detached at 8.7.65))$ module --version

Modules based on Lua: Version 8.7.65 2025-08-05 10:24 -06:00
    by Robert McLay mclay@tacc.utexas.edu
```

smoors:
i finally found what's causing this in our site: it's our `lmod_config.lua`:

```lua
require("strict")
local cosmic = require("Cosmic"):singleton()
cosmic:assign("LMOD_ADMIN_FILE", "/etc/lmod/admin.list")
cosmic:assign("LMOD_AUTO_SWAP", "no")
cosmic:assign("LMOD_CACHED_LOADS", "yes")
cosmic:assign("LMOD_CASE_INDEPENDENT_SORTING", "yes")
cosmic:assign("LMOD_DISABLE_SAME_NAME_AUTOSWAP", "yes")
cosmic:assign("LMOD_EXTENDED_DEFAULT", "no")
cosmic:assign("LMOD_PACKAGE_PATH", "/etc/lmod")
cosmic:assign("LMOD_PIN_VERSIONS", "yes")
cosmic:assign("LMOD_REDIRECT", "yes")
cosmic:assign("LMOD_SHORT_TIME", 86400)
cosmic:assign("LMOD_SITE_MSG_FILE", "/etc/lmod/lang.lua")
cosmic:assign("LMOD_SITE_NAME", "VUB_HPC")
```

if i comment out the line with `LMOD_CACHED_LOADS`, it works, but strangely enough just `export LMOD_CACHED_LOADS=no` does not override it.

smoors:
@ocaisa can you try to repro with `export LMOD_CACHED_LOADS=yes`?

ocaisa:
Similar behaviour for me (in a way), if I export that variable it seems to have no effect and I can still successfully load the module

ocaisa:
No, sorry, I am wrong, this does indeed reproduce the problem
```
ocaisa@~/Lmod((HEAD detached at 8.7.65))$ export LMOD_CACHED_LOADS=yes
ocaisa@~/Lmod((HEAD detached at 8.7.65))$ module purge
ocaisa@~/Lmod((HEAD detached at 8.7.65))$ module load EESSI/2025.06
Lmod has detected the following error:  Unable to load module because of error when evaluating modulefile:
     /cvmfs/software.eessi.io/init/modules/EESSI/2025.06.lua: /home/ocaisa/lmod/lmod/lmod/libexec/ModuleA.lua:676: Did not find mpath in mA

     Please check the modulefile and especially if there is a line number specified in the above message
While processing the following module(s):
    Module fullname  Module Filename
    ---------------  ---------------
    EESSI/2025.06    /cvmfs/software.eessi.io/init/modules/EESSI/2025.06.lua
```

boegel:
@ocaisa We also have cached loads enabled on the HPC-UGent infrastructure, just in case that's helpful to debug...

smoors:
> @ocaisa We also have cached loads enabled on the HPC-UGent infrastructure, just in case that's helpful to debug...

but the version in your systems is not affected, only 8.7.65 to 9.0.4

ocaisa:
`haveDynamicMPATH()` does not seem to help

casparvl:
We could just document this as a known issue for versions X-Y, when `LMOD_CACHED_LOADS=yes` is configured. And then advise that for these versions, people use `module load --ignore_cache EESSI/<version>` (instead of setting `LMOD_IGNORE_CACHE`, which could easily be exported 'by mistake'). And stress that the `--ignore_cache` argument should _only_ be passed when loading the EESSI module - not when loading any subsequent module.

Even better would be if we can detect the LMOD version and `LMOD_CACHE_LOADS` setting in the EESSI module - and print helpful advice from there. But I don't know if that's possible/easy.

smoors:
unfortunately, `module --ignore_cache load EESSI/<version>` doesn't work..

casparvl:
> unfortunately, module --ignore_cache load EESSI/<version> doesn't work..

That's... baffling. I would have assumed it to be _identical_ to `LMOD_IGNORE_CACHE` :\

ocaisa:
Is there something else special about your Lmod setup? It is indeed pretty strange that the cached loads is not affected by the envvar that is supposed to affect it, and the command line option equivalent of `LMOD_IGNORE_CACHE` also does not work.

smoors:
for `LMOD_CACHED_LOADS`, i think the `lmod_config.lua` file takes precedence over the envvar.

for `--ignore_cache`, it definitely works when loading "normal" modules, so i suspect this is related to the magic that is happening in the EESSI module?

anyway, it looks like an unhappy coincidence of a lot of factors, so maybe not worth adding this to the docs?

ocaisa:
Perhaps a FAQ entry...if we had such a thing

PR #734, opened 2026-04-10
Update supported CPU/GPU targets for 2025.06

bedroge:
CPU overview:

image

GPU overview:

image

Not completely happy with this one yet. Already disabled the TOC to get more space, but there's quite a lot of information in the table. Maybe we should shorten the CPU names here or stick to the same ones as on the CPU page (i.e. `x86_64/amd/zen5` instead of having different columns)?

But, anyway, I'd say it's already an improvement (currently there's no info about 2025.06 at all), so we can further improve it in follow-up PRs.

### DIFF
--- a/docs/software_layer/gpu_targets.md
+++ b/docs/software_layer/gpu_targets.md
@@ -1,130 +1,175 @@
+---
+hide:
+  - toc
+---
+
 # GPU targets
 
 ## NVIDIA
 
-EESSI includes optimized software installations for major NVIDIA GPU CUDA Compute Capabilities 7.0, 8.0 and 9.0, for all CPU targets listed in [CPU targets](cpu_targets.md).
+Each version of EESSI includes optimized software installations for a selected range of major NVIDIA GPU CUDA Compute Capabilities,
+for all CPU targets listed in [CPU targets](cpu_targets.md).
 
-NVIDIA guarantees forwards compatibility within a major CUDA Compute Capability (i.e. a GPU with CUDA Compute Capability 8.6 can run code built for CUDA Compute Capability 8.0).
-This means that all NVIDIA GPUs with Compute Capability 7.x, 8.x and 9.x are supported (see also [here](https://developer.nvidia.com/cuda-gpus)).
+NVIDIA guarantees forward compatibility within a major CUDA Compute Capability, i.e. a GPU with CUDA Compute Capability 8.6 can run code built for CUDA Compute Capability 8.0.
+This means that a given EESSI version supports all NVIDIA GPUs with Compute Capability X.Y if the table below includes support for Compute Capability X.0.
+On the [NVIDIA website](https://developer.nvidia.com/cuda/gpus) you can find the Compute Capability of your GPU.
 
-The decision to only ship code optimized for the major GPU architectures was made to keep the amount of software builds that need to be done at a reasonable level.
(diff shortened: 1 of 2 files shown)


bedroge:
I've collapsed the CPU target columns into a single one, I think it looks better now:

image

docs/software_layer/gpu_targets.md
```diff
@@ -1,130 +1,170 @@
...
         </tr>
     </thead>
     <tbody>
         <tr>
-            <td rowspan=4>aarch64</td>
-            <td colspan=2>generic</td>
-            <td>x</td>
-            <td>x</td>
-            <td>x</td>
+            <td><code>aarch64/generic`</code></td>
```
casparvl:
```suggestion
            <td><code>aarch64/generic</code></td>
```

docs/software_layer/gpu_targets.md
```diff
@@ -1,130 +1,170 @@
...
+NVIDIA guarantees forward compatibility within a major CUDA Compute Capability, i.e. a GPU with CUDA Compute Capability 8.6 can run code built for CUDA Compute Capability 8.0.
+This means that a given EESSI version supports all NVIDIA GPUs with Compute Capability X.Y if the table below includes support for Compute Capability X.0.
+On the [NVIDIA website](https://developer.nvidia.com/cuda/gpus) you can find the Compute Capability of your GPU.
 
-The decision to only ship code optimized for the major GPU architectures was made to keep the amount of software builds that need to be done at a reasonable level.
-Even just building for the major CUDA Compute Capabilities for all CPU targets currently requires 39 builds (13 CPU targets times 3 GPU targets).
+The decision to only ship code optimized for the major GPU architectures was made to keep the number of software builds that need to be done at a reasonable level.
+Even just building for three major CUDA Compute Capabilities for all CPU targets supported by EESSI version 2023.06 would require 39 builds in total (13 CPU targets times 3 GPU targets).
```
casparvl:
```suggestion
The decision to only ship code optimized for the major GPU architectures was made to keep the number of software builds that need to be done at a reasonable level.
Even just building for five major CUDA Compute Capabilities for all CPU targets supported by EESSI version 2025.06 already requires 60 builds in total (15 CPU targets times 5 GPU targets).
```

bedroge: 15*5 != 60 :stuck_out_tongue:

bedroge: fixed in https://github.com/EESSI/docs/pull/734/commits/ca3074ad92870146fcd52f0dc74826a3cc4e594a

bedroge:
I've replaced all the `<span class="software-eessi-version-XXXX.06">XXXX.06</span>` elements by variables that are defined at the op of the pages. This makes the markdown/HTML code for the tables a bit cleaner/shorter.

bedroge:
Also disabled the TOC for the CPU page, resulting in a wider table (and the long CPU target names are no longer split into two lines):
image

And, just to ensure that things still render correctly with the new variables, this is the GPU page:
image

casparvl (approved):
Looks good to me. Thanks Bob, it's been a thorn in my side that our efforts on the 2025.06 GPU support weren't "visible" yet - now they are. And it looks a lot more attractive visually than it used to. Nice!

PR #750, opened 2026-04-29
Use correct diacritic

Neves-P:
No nasal sounds here :smile:


ocaisa (approved):
I would like to hear what this would sound like

bedroge (approved):
Great improvement, best PR ever 😄

PR #778, opened 2026-06-04
Document approaches for site builds on top of EESSI

### LINKED ISSUE #768: Add documentation on how to build on top of EESSI as a site - and include the use of `EESSI_SITE_SOFTWARE_PREFIX` in that
We should have documentation on our recommended ways to build on top of EESSI as a site. One thing that should be covered is how to use `EESSI_SITE_SOFTWARE_PREFIX` to control the location of where these installations end up (this was introdcued in https://github.com/EESSI/software-layer-scripts/pull/232).

casparvl:


docs/site_build/overview.md
```diff
@@ -0,0 +1,45 @@
...
+1. You want to offer software that does is not suitable for upstream deployment in EESSI (e.g. because it is proprietary, or because it is a development build / otherwise very specific build that is not useful for a general audience).
+2. You need to make software available on (very) short notice to your users, and cannot wait for it to be deployed in upstream EESSI.
+3. You want to retain full autonomy over what gets deployed
+
+While all of these are valid arguments, note that there is also one major downside to deploying things locally: you loose one of the core benefits of EESSI, namely that it provides _the same software on every system_. The more site-specific installations you have, the more difficult it will be for your users to move their workflows from e.g. their own development machine/cloud environment to your cluster, or scale up to larger clusters. If you're doing site-builds to make software available to your users on short notice, we highly encourage you to _also_ contribute the same software installation in upstream EESSI. This way, once accepted upstream, users that rely on that software retain their 'mobility'.
```
ocaisa:
```suggestion
While all of these are valid arguments, note that there is also one major downside to deploying things locally: you loose one of the core benefits of EESSI, namely that it provides _the same software on every system_. The more site-specific installations you have, the more difficult it will be for your users to move their workflows from, e.g., their own development machine/cloud environment to your cluster, or scale up to larger clusters. If you're doing site-builds to make software available to your users on short notice, we highly encourage you to _also_ contribute the same software installation in upstream EESSI. This way, once accepted upstream, users that rely on that software retain their 'mobility'.
```

docs/site_build/overview.md
```diff
@@ -0,0 +1,45 @@
...
+
+# Choosing your approach
+There are two approaches to doing site builds, each with their own advantages and disadvantages.
+
+1. Perform site builds using EESSI-extend on a shared filesystem.
+2. Leverage EESSI's build procedure for site builds. In this approach, you use the EESSI build bot (`EESSI/eessi-bot-software-layer`), together with the EESSI build scripts (`EESSI/software-layer-scripts`) to build and deploy software into a CernVM-FS repository of your own. Essentially, this means you'll build in a way that is essentially identical to how it is done for upstream EESSI - with the only major difference being the target CernVM-FS repository.
+
+In both cases, you build 'on top' of EESSI, meaning that dependencies that are already provided by EESSI will not be reinstalled: they will simply be loaded from EESSI.
+
+Here, we list some advantages and disadvantages to help you choose which approach best suites your requirements.
```
ocaisa:
```suggestion
Here, we list some advantages and disadvantages to help you choose which approach best suits your requirements.
```

docs/site_build/overview.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Approach 1: using EESSI-extend on shared FS
+
+Advantages:
+- Easy to get started: no additional setup or knowledge needed
+- Automatically optimizes for the host on which you run the installation, and installs in architecture-specific prefix that matches the host architecture. This means you can install optimized software for each of your CPU/GPU architectures in an organized way.
+
+Disadvantages:
+- This is a manual procedure (unless you create your own automation around it). As such, doesn't scale well to installing large amounts of software and/or installing software for many different hardware targets.
```
ocaisa:
```suggestion
- This is a manual procedure (unless you create your own automation around it). As such, it doesn't scale well to installing large amounts of software and/or installing software for many different hardware targets.
```

docs/site_build/overview.md
```diff
@@ -0,0 +1,45 @@
...
+- The fact that you get optimized installations means that on a very heterogeneous system, you will have to run the installation many times - once for each architecture on which you want to offer that particular piece of software.
+- Shared filesystems (and especially _parallal_ filesystems) are generally ill-suited to serve software. This means start-up time can be quite long (you can find some numbers [here](../training-events/2025/tutorial-best-practices-cvmfs-hpc/performance.md)).
```
ocaisa:
```suggestion
- Shared filesystems (and especially _parallel_ filesystems) are generally ill-suited to serve software. This means start-up time can be quite long (you can find some numbers [here](../training-events/2025/tutorial-best-practices-cvmfs-hpc/performance.md)).
```

docs/site_build/overview.md
```diff
@@ -0,0 +1,45 @@
...
+Advantages:
+- Highly automated
+- Scalable to many architectures & installations
+- Site builds are done based on a list of software in a GitHub repo - making it very transparent what is available / got added on your system
+- Share maintenance on the automation with the EESSI community
+- End-user look & feel are very similar to EESSI
+
+Disadvantages
+- More setup time
+- Requires more extnesive knowledge (CVMFS, EESSI build bot, object store)
```
ocaisa: Why is codespell not catching this?
```suggestion
- Requires more extensive knowledge (CVMFS, EESSI build bot, object store)
```

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,221 @@
+# Leverage EESSI's build procedure for site builds
+In this approach, you use the EESSI build bot (`EESSI/eessi-bot-software-layer`), together with the EESSI build scripts (`EESSI/software-layer-scripts`) to build and deploy software into a CernVM-FS repository of your own. Essentially, this means you'll build in a way that is essentially identical to how it is done for upstream EESSI - with the only major difference being the target CernVM-FS repository.
```
ocaisa:
```suggestion
In this approach, you use the EESSI build bot (`EESSI/eessi-bot-software-layer`), together with the EESSI build scripts (`EESSI/software-layer-scripts`) to build and deploy software into a CernVM-FS repository of your own. Essentially, this means you'll build in a way that is effectively identical to how it is done for upstream EESSI - with the only major difference being the target CernVM-FS repository.
```

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,221 @@
...
+
+!!! note
+
+    The recommended CVMFS setup requires a fair amount of machines. If this is more than you can afford, there are some tricks you can pull. First, you can combine each proxy with a Stratum 1 on the same machine, only use the proxies for proxy-ing upstream EESSI, and simply have your clients contact your site-specific Stratum 1's directly (without proxy). In this scenario, you can achieve load-balancing by configuring half your clients with `CVMFS_SERVER_URL="<instance_1>;<instance_2>"` and half with `CVMFS_SERVER_URL="<instance_2>;<instance_1>"`, where `instance_1` and `instance_2` are the IPs of your Stratum 1's. Finally, you can even use the Stratum 0 instead of a second Stratum 1. Note that this has security implications, as it means your Stratum 0 needs to be directly accessible to your clients. This is a potential concern: if there are vulnarebilities in the Stratum 0 software, end-users may be able to push (malicious) software in there.
```
ocaisa:
```suggestion
    The recommended CVMFS setup requires a fair amount of machines. If this is more than you can afford, there are some tricks you can pull. First, you can combine each proxy with a Stratum 1 on the same machine, only use the proxies for proxy-ing upstream EESSI, and simply have your clients contact your site-specific Stratum 1's directly (without proxy). In this scenario, you can achieve load-balancing by configuring half your clients with `CVMFS_SERVER_URL="<instance_1>;<instance_2>"` and half with `CVMFS_SERVER_URL="<instance_2>;<instance_1>"`, where `instance_1` and `instance_2` are the IPs of your Stratum 1's. Finally, you can even use the Stratum 0 instead of a second Stratum 1. Note that this has security implications, as it means your Stratum 0 needs to be directly accessible to your clients. This is a potential concern: if there are vulnerabilities in the Stratum 0 software, end-users may be able to push (malicious) software in there.
```

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,221 @@
...
+```
+
+5. Create the repository owned by `root`:
+```bash
+sudo cvmfs_server mkfs -o root $repo_name
+```
+
+!!! note
+
+    The reason we configure `root` to be the owner of the CVMFS repository is that EasyBuild, when configured through `EESSI-extend`, by default creates read-only installations. This causes issues if CVMFS has to put catalog files (`.cvmfscatalog`) files in these directories, which are metadata files that CVMFS uses to list the files/directories present in the repository. While it is technically possible to use a regular user, this would require making all directories in which CVMFS would create a `.cvmfscatalog` file writeable in a transaction, then create the catalog files, then remove the write permissions again. The same approach would need to be taken to reinstall software that was already installed. We consider this unnecessarily complex, and instead prefer to have the repository owned by root.
```
ocaisa:
```suggestion
    The reason we configure `root` to be the owner of the CVMFS repository is that EasyBuild, when configured through `EESSI-extend`, by default creates read-only installations. This causes issues if CVMFS has to put catalog files (`.cvmfscatalog`) files in these directories, which are metadata files that CVMFS uses to list the files/directories present in the repository. While it is technically possible to use a regular user, this would require making all directories in which CVMFS would create a `.cvmfscatalog` file writeable in a transaction, then create the catalog files, then remove the write permissions again. The same approach would need to be taken to reinstall software what was already installed. We consider this unnecessarily complex, and instead prefer to have the repository owned by root.
```

ocaisa (changes requested):
Found some typos

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,420 @@
...
+- An instance of the EESSI build bot
+- A bucket in an AWS S3-compatible object store (though you could work around this)
+- A GitHub organization on which you can install GitHub Apps
+- A GitHub repository within that organization which will be used to list the software you want to build
+- Optionally: an automated procedure to ingest tarballs
+
+This documentation will go through the steps to set each of these up, in order. Since many of these individual steps are documented elsewhere, we will often reference that (and only list a very short summary here).
+
+## Site-specific CVMFS infrastructure
+The recommended CVMFS setup for a site-specific CVMFS repository is:
```
casparvl, line 17: TODO: Add picture for this

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1451 @@
...
+aws s3 mb s3://<bucket_name>
+```
+
+### Security considerations
+
+There are many policies you can attach to a bucket to increase security. At the _very_ least, consider setting a bucket policy that restricts IP access to a whitelisted range (i.e. only the nodes where your build bot(s) run, and your Stratum 0 need access). The [AWS policy generator](https://awspolicygen.s3.amazonaws.com/policygen.html) can be a helpful tool in generating such a policy.
+
+Another thing to consider is to create a secondary user with `aws iam create-user --user-name <name>` and attach a very limited policy to it (e.g. only read/write/list on buckets, nothing else). Then, create credentials for this user with `aws iam create-access-key --user-name <name>` and provide those credentials to the EESSI build bot and Stratum 0 machines. That way, if that token is compromized, the impact is minimized (e.g. the token can at least not be used to create new IAM idententies, etc).
+
+## Setting up the EESSI build bot
```
Neves-P, line 850: I'd consider separating this into a new page. The build bot is somewhat self contained and the page is quite long already.

docs/site_build/site_cvmfs.md
casparvl: Err, are you sure? "reinstall software _what_ was already installed"? That sounds wrong to me - but might be my non-native ears :)

docs/site_build/shared_fs.md
```diff
@@ -0,0 +1,137 @@
...
+fail-on-mod-files-gcccore                (E) = True
+filter-deps                              (E) = binutils, bzip2, DBus, flex, gettext, gperf, help2man, intltool, libreadline, makeinfo, ncurses, NVPL, ParMETIS, util-linux, XZ, zlib
+filter-env-vars                          (E) = LD_LIBRARY_PATH
+hooks                                    (E) = /cvmfs/software.eessi.io/versions/2025.06/init/easybuild/eb_hooks.py
+ignore-osdeps                            (E) = True
+installpath                              (E) = /sharedfs/eessi/versions/2025.06/software/linux/x86_64/amd/zen3
+...
+```
+
+Nowm, we can start building, e.g.:
```
Neves-P:
```suggestion
Now, we can start building, e.g.:
```

docs/site_build/shared_fs.md
```diff
@@ -0,0 +1,137 @@
...
+
+When the installation has completed, the software should be available in your prefix.
+
+## Container
+
+Instead of doing the builds on the host system itself, you could consider doing them in a container.
+Using a minimal build container minimizes the risk of accidentally picking up host libraries (instead of the ones provided by EESSI),
+and a container also provides a controlled and isolated environment.
+
+In principle you could any container, as long as you make sure that both the EESSI CVMFS repository and your shared file system are available in the container.
```
Neves-P:
```suggestion
In principle you could use any container, as long as you make sure that both the EESSI CVMFS repository and your shared file system are available in the container.
```

docs/site_build/shared_fs.md
```diff
@@ -0,0 +1,137 @@
...
+Assuming both are available on the build host, you can simply bind mount both of them into the container.
+
+You can also use the (`eessi_container.sh` script)[https://github.com/EESSI/software-layer-scripts/blob/main/eessi_container.sh], provided by EESSI,
```
Neves-P:
```suggestion
You can also use the [`eessi_container.sh` script](https://github.com/EESSI/software-layer-scripts/blob/main/eessi_container.sh), provided by EESSI,
```

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1708 @@
...
+proxy_ip2=<IP_OR_DNS_NAME_OF_PROXY2>
+proxy_port2=<PORT_NR_FOR_PROXY2>
+```
+
+**2. Install CVMFS client**
+
+Typically, the machines on which you want to offer your own software stack on top of EESSI already have the CVMFS client installed, otherwise you wouldn't be able to serve EESSI there. If you haven't done so, please follow the instructions [here](../../getting_access/native_installation#native-install-on-clusters).
+
+**3. Add the repository master public key**
+On your CVMFS **Stratum 0**, check the contents of your master key:
```
Neves-P, line 612:
```suggestion

On your CVMFS **Stratum 0**, check the contents of your master key:
```

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1708 @@
...
+```
+cat "/etc/cvmfs/keys/${repo_name}.pub"
+```
+and copy that to /etc/cvmfs/keys/${site_tld}/${repo_name}.pub on your client machines.
```
Neves-P:
```suggestion
and copy that to `/etc/cvmfs/keys/${site_tld}/${repo_name}.pub` on your client machines.
```

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1708 @@
...
+```
+
+Note that we assumed you have signature files as well. If not, remove those sections from the code above.
+
+**Full script**
+
+Composing all of the above (including signature verification), we get the following script for automatic ingestion:
+
+{% raw %}
+```
```
Neves-P:
```suggestion
``` { .bash .copy }
```

Neves-P (changes requested):
Looks really nice! This is my first pass of comments, as of know I only focused on the text. This week I will follow the documentation in practice and review again.

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1709 @@
...
+aws s3 mb s3://<bucket_name>
+```
+
+### Security considerations
+
+There are many policies you can attach to a bucket to increase security. At the _very_ least, consider setting a bucket policy that restricts IP access to a whitelisted range (i.e. only the nodes where your build bot(s) run and your Stratum 0 need access). The [AWS policy generator](https://awspolicygen.s3.amazonaws.com/policygen.html) can be a helpful tool in generating such a policy.
+
+Another thing to consider is to create a secondary user with `aws iam create-user --user-name <name>` and attach a very limited policy to it (e.g. only read/write/list on buckets, nothing else). Then, create credentials for this user with `aws iam create-access-key --user-name <name>` and provide those credentials to the EESSI build bot and Stratum 0 machines. That way, if that token is compromised, the impact is minimized (e.g. the token can at least not be used to create new IAM idententies, etc).
+
+## Setting up the EESSI build bot
```
casparvl, line 850: I realize these docs lack one crucial thing: setting up a `site_config_script` that sets `EESSI_SITE_INSTALL_FORCE` and `EESSI_SITE_SOFTWARE_PATH_PREFIX`

casparvl: I.e. at SURF, our `site_config_script` looks like this:

```
# To build on top of EESSI, we need to software.eessi.io repository to be mounted next to our own repository
# The bot/build.sh script does this when the EESSI_SITE_INSTALL_FORCE environment variable is set
# Other build scripts will also respect this variable where needed in order to make sure that 'building on top'
# of EESSI is possible
export EESSI_SITE_INSTALL_FORCE=1
echo "Value of EESSI_SITE_INSTALL_FORCE: $EESSI_SITE_INSTALL_FORCE"

# We also need to set a prefix that our installations should end up in
# The build scripts should take this prefix, and construct the final EESSI_SITE_SOFTWARE_PATH out of it
# that the EESSI-extend module expects
export EESSI_SITE_SOFTWARE_PATH_PREFIX=/cvmfs/software.surf.nl/versions/2025.06/
echo "Value of EESSI_SITE_SOFTWARE_PATH_PREFIX: $EESSI_SITE_SOFTWARE_PATH_PREFIX"
```

One thing to figure out might be how to avoid hard-coding the version in here :)

casparvl: Scratch that, this should not be needed anymore... I think we do this differently now... We automatically set these here https://github.com/EESSI/software-layer-scripts/blob/ca929cd7ef32a9fcd79fafd4e0d5c362a1fff452/bot/build.sh#L156 . That happens if the `repo_name` in the repos config file is unequal to `software.eessi.io` or `dev.eess.io` essentially.

Neves-P: I think this does warrant a small explanation of why `EESSI_SITE_SOFTWARE_PREFIX`, `EESSI_SITE_INSTALL` and friends are not needed with this workflow. Maybe we don't even need an explanation, but just a mention. 

I'm saying this because going through the whole docs - especially the shared file system installs - you get the impression that at least doing `EESSI_SITE_INSTALL` is necessary for site installations (which it is, but that happens behind the scenes).

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1709 @@
...
+There is extensive documentation on how to interact with S3 buckets available online - here we only list the few commands that you would commonly need to set up a staging bucket. Note that you can create the bucket from anywhere - but you'll need the AWS commands to be available on the machine where you [install the EESSI build bot](#install-eessi-build-bot-on-a-machine), so you might as well set it up there directly.
+
+### Installing the AWS CLI commands
+
+Today, this can be done with:
+
+``` { .bash .copy }
+curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
+unzip awscliv2.zip
+sudo ./aws/install
```
Neves-P, line 792: This can be installed in userland, and ideally I'd do so. That's how I normally install it in the `dev.eessi.io` bots and it works well.

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1709 @@
...
+
+## Setting up an object store to stage build tarballs
+
+The standard deployment method for the EESSI build bot is to stage tarballs in an S3 bucket. While the bot's functionality may be extended in the future (the [function](https://github.com/EESSI/eessi-bot-software-layer/blob/29dc5e9aa339c323a900dc1454d39246def73984/tasks/deploy.py#L689) actually uploading the tarballs could easily be altered to deploy to a central directory on a local filesystem, for example), for now this means we need an S3 bucket if we want to use the bot's deployment functionality.
+
+There is extensive documentation on how to interact with S3 buckets available online - here we only list the few commands that you would commonly need to set up a staging bucket. Note that you can create the bucket from anywhere - but you'll need the AWS commands to be available on the machine where you [install the EESSI build bot](#install-eessi-build-bot-on-a-machine), so you might as well set it up there directly.
```
Neves-P, line 783:
```suggestion
There is extensive documentation on how to interact with S3 buckets available online - here we only list the few commands that you would commonly need to set up a staging bucket. Note that you can create the bucket from anywhere - but you'll need the AWS commands to be available on the machine where you [install the EESSI build bot](#install-eessi-build-bot-on-a-machine), so you might as well set it up there directly. You don't necessarily need to use Amazon AWS S3 buckets, for example, if you have access to object storage on a Ceph cluster. The only requirement is that the `aws` tool can read and write to your endpoint.
```

Neves-P (changes requested):
Just two comments, suggestions. I would have liked to have been more thorough, but overall I did two passes to the text and managed to get to the point of installing the bot (had to wait a bit for the internal S3 and maintenance).

In my mind, after addressing these comments we can either wait a little longer and I spin up a new test CVMFS infrastructure to also test that part, or merge before that and address any thing else we spot in a later PR.

Overall I think the instructions are very nice! I'm not sure I am able to put myself in the shoes of a complete novice to EESSI, but they seem very appropriate for that target audience.

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1709 @@
...
+- Under "Subscribe to events", tick the "Issue comment" and "Pull request" boxes (note: these only appear once you've set the Repository Permissions above)
+- Select "Only allow this GitHub App to be installed on the GH_ORG account
+- Click "Create GitHub App"
+- Scroll down to the "Private keys" section and click "Generate a private key". This key will be used by the bot event handler in order to be able to interact with GitHub (and e.g. post responses in your PRs).
+
+### Create a GitHub repository
+
+- Go to https://github.com/organizations/GH_ORG/repositories/new
+- Pick a descriptive name. We'll refer to it as GH_REPO
+- Add the `bot/build.sh` from [EESSI/software-layer](https://github.com/EESSI/software-layer) to your repository (under the exact same name)
```
Neves-P, line 905:
```suggestion
- Add the `bot/build.sh` from [EESSI/software-layer](https://github.com/EESSI/software-layer) to your repository (under the exact same name) and make sure it is executable by the Linux user running the bot.
```

Neves-P (changes requested):
Another small detail that I spotted now and am adding here so I don't forget.

Neves-P:
The last commit message was supposed to read: "Account for custom prefix installs for .cvmfsdirtab"

After that is reviewed, we need to account for the changes in the tarball ingestion step too. That is blocked until https://github.com/EESSI/filesystem-layer/pull/278 is merged.

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1761 @@
...
+
+``` { .bash .copy }
+# Repository-relative path to the versions directory (default)
+versions_subpath=versions
+# Alternatively, define a different versions subpath that includes
+# an extra subdirectory
+# versions_subpath=eessi/versions
+```
+
+This is especially useful if you want your repository to have custom site software prefix. While a software prefix looks like `/cvmfs/${repo_name}/`, you might want to, for example, use `/cvmfs/${repo_name}/eessi/`. If you do so, you will also need to use  `EESSI_SITE_SOFTWARE_PREFIX` later, more details on that later in the [bot configuration section](#install-eessi-build-bot-on-a-machine).
```
bedroge:
```suggestion
This is especially useful if you want your repository to have a custom site software prefix. While a default software prefix looks like `/cvmfs/${repo_name}/`, you might want to, for example, use `/cvmfs/${repo_name}/eessi/`. If you do so, you will also need to use  `EESSI_SITE_SOFTWARE_PREFIX` later, more details on that later in the [bot configuration section](#install-eessi-build-bot-on-a-machine).
```

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1761 @@
...
+
+```
+aws s3 mb s3://<bucket_name>
+```
+
+### Security considerations
+
+There are many policies you can attach to a bucket to increase security. At the _very_ least, consider setting a bucket policy that restricts IP access to a whitelisted range (i.e. only the nodes where your build bot(s) run and your Stratum 0 need access). The [AWS policy generator](https://awspolicygen.s3.amazonaws.com/policygen.html) can be a helpful tool in generating such a policy.
+
+Another thing to consider is to create a secondary user with `aws iam create-user --user-name <name>` and attach a very limited policy to it (e.g. only read/write/list on buckets, nothing else). Then, create credentials for this user with `aws iam create-access-key --user-name <name>` and provide those credentials to the EESSI build bot and Stratum 0 machines. That way, if that token is compromised, the impact is minimized (e.g. the token can at least not be used to create new IAM idententies, etc).
```
bedroge:
```suggestion
Another thing to consider is to create a secondary user with `aws iam create-user --user-name <name>` and attach a very limited policy to it (e.g. only read/write/list on buckets, nothing else). Then, create credentials for this user with `aws iam create-access-key --user-name <name>` and provide those credentials to the EESSI build bot and Stratum 0 machines. That way, if that token is compromised, the impact is minimized (e.g. the token can at least not be used to create new IAM identities, etc).
```

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1761 @@
...
+
+``` { .bash .copy }
+#!/bin/bash
+
+IFS=$'\n\t'         # sane field splitting
+BUCKET="<bucket_name>"
+DOWNLOAD_DIR=/prefix/for/tarball/staging  # Some directory to temporarily store tarballs on the Stratum 0
+ALLOWED_SIGNERS=/path/to/allowed/signers/file  # Optional, needed in step 4
+REPO_NAME="<repo_name>"
+# a name for a dir in the bucket in which to archive tarballs, so that a subsequent run doesn't re-ingest them
```
Neves-P, line 1203:
```suggestion
# Repository-relative path to the versions directory.
# This must match the versions_subpath used for .cvmfsdirtab.
VERSIONS_SUBPATH="versions"
# For a site prefix such as /cvmfs/name.sitename.tld/eessi, use:
# VERSIONS_SUBPATH="eessi/versions"

# a name for a dir in the bucket in which to archive tarballs, so that a subsequent run doesn't re-ingest them
```

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1761 @@
...
+# We will leverage a script from filesystem-layer (for tarball ingestion)
+git clone https://github.com/EESSI/filesystem-layer.git
+
+# The script referenced here does three things
+# 1. Ingest the tarball (cvmfs_server ingest <tarball>)
+# 2. Regenerate the `.cvmfscatalog` files by publishing an empty transaction (`cvmfs_server transaction && cvmfs_server -m "<commitg_msg>"`)
+# 3. Update the lmod caches for your site installation prefix
+INGEST_SCRIPT="$PWD/filesystem-layer/scripts/ingest-tarball.sh"
+```
+
```
Neves-P, line 1220:
```suggestion

`VERSIONS_SUBPATH` must match the `versions_subpath` used when [configuring `.cvmfsdirtab`](#setting-up-your-stratum-0). It is relative to the root of the target CVMFS repository and must include the `versions` directory.

For example, with `EESSI_SITE_SOFTWARE_PREFIX=/cvmfs/name.sitename.tld/eessi`, use `VERSIONS_SUBPATH=eessi/versions`. The default is `versions`.

```

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1761 @@
...
+        rm -f "${local_tar}" "${local_meta}" "${local_tar_sig}" "${local_meta_sig}"
+        continue
+    fi
+```
+
+Note that our `rm -f` assumes you downloaded signature files (`${local_tar_sig}` and `${local_meta_sig}`) as well - if not, you'll have to strip that from the command.
+
+**5. Ingest the tarball into the repository**
+
+Here, we leverage a script from `EESSI/filesystem-layer` that ingests tarballs, but also takes care of regenerating the `.cvmfscatalog` files _and_ updates the `Lmod` cache. To update the `Lmod` cache, the script uses the Lmod installation provided by `software.eessi.io`, which is why we explicitly made this available as one of the steps [when we set up our Stratum 0](#setting-up-your-stratum-0)
```
Neves-P, line 1365:
```suggestion
Here, we leverage [`ingest-tarball.sh`](https://github.com/EESSI/filesystem-layer/blob/main/scripts/ingest-tarball.sh) from `EESSI/filesystem-layer`. The script ingests the tarball, regenerates the `.cvmfscatalog` files, and updates the Lmod cache.

The tarball contains paths relative to the versioned installation directory, such as `2025.06/software/...`. The `--basedir` option determines where that directory is placed relative to the root of the target CVMFS repository. It must therefore match the repository layout configured through `VERSIONS_SUBPATH`.

When updating the Lmod cache, the script uses the selected base directory for the site installation. If the site repository does not provide its own compatibility layer, it automatically falls back to an Lmod installation from `/cvmfs/software.eessi.io/versions`.
```

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1761 @@
...
+``` { .bash .copy }
+    # ---- Ingest into CVMFS ----
+    echo "Ingesting into CVMFS (${REPO_NAME})..."
+    if $INGEST_SCRIPT "${REPO_NAME}" "${local_tar}"; then
+        echo "Ingest succeeded for ${filename}."
+    else
+        echo "ERROR: cvmfs_server ingest failed for ${filename}." >&2
+        # Keep the files for troubleshooting
+        continue
+    fi
```
Neves-P, line 1376:
```suggestion
    echo "Ingesting into CVMFS (${REPO_NAME}) under ${VERSIONS_SUBPATH}..."
    if "${INGEST_SCRIPT}" \
        --repository "${REPO_NAME}" \
        --basedir "${VERSIONS_SUBPATH}" \
        "${local_tar}"; then
        echo "Ingest succeeded for ${filename}."
    else
        echo "ERROR: CVMFS ingestion failed for ${filename}." >&2
        # Keep the files for troubleshooting
        continue
    fi
```

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1761 @@
...
+```
+
+**6. Regenerate the `.cvmfscatalog` files by publishing an empty transaction**
+
+This is already taken care of by the `$INGEST_SCRIPT` in the step above. If you don't want to use that script, you'll have to implement this step yourself.
```
Neves-P, line 1381:
```suggestion
This is already taken care of by `$INGEST_SCRIPT` in the step above. The patterns in `.cvmfsdirtab` must use the same repository-relative versions path as `VERSIONS_SUBPATH`; otherwise the expected nested catalogs will not be created.
```

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1761 @@
...
+
+**7. Open a new transaction, update the Lmod cache for your site installs, and publish the transaction**
+
+This is already taken care of by the `$INGEST_SCRIPT` in step 5 above. If you don't want to use that script, you'll have to implement this step yourself.
```
Neves-P, line 1385:
```suggestion
This is already taken care of by `$INGEST_SCRIPT` in step 5. The cache is updated below `${VERSIONS_SUBPATH}/<eessi_version>` in the site repository. The script will use an Lmod installation from the site repository when one is available, and otherwise falls back to `software.eessi.io`.

```

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1761 @@
...
+
+{% raw %}
+``` { .bash .copy }
+#!/bin/bash
+
+IFS=$'\n\t'         # sane field splitting
+BUCKET="<bucket_name>"
+DOWNLOAD_DIR=/prefix/for/tarball/staging  # Some directory to temporarily store tarballs on the Stratum 0
+ALLOWED_SIGNERS=/path/to/allowed/signers/file  # Optional, needed in step 4
+REPO_NAME="<repo_name>"
```
Neves-P, line 1460:
```suggestion
REPO_NAME="<repo_name>"

# Repository-relative path to the versions directory.
# This must match the versions_subpath used for .cvmfsdirtab.
VERSIONS_SUBPATH="versions"
# For a site prefix such as /cvmfs/name.sitename.tld/eessi, use:
# VERSIONS_SUBPATH="eessi/versions"

```

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1761 @@
...
+
+    # ---- Ingest into CVMFS ----
+    echo "Ingesting into CVMFS (${REPO_NAME})..."
+    if $INGEST_SCRIPT "${REPO_NAME}" "${local_tar}"; then
+        echo "Ingest succeeded for ${filename}."
+    else
+        echo "ERROR: cvmfs_server ingest failed for ${filename}." >&2
+        # Keep the files for troubleshooting
+        continue
+    fi
```
Neves-P, line 1576:
```suggestion
    # ---- Ingest into CVMFS ----
    echo "Ingesting into CVMFS (${REPO_NAME}) under ${VERSIONS_SUBPATH}..."
    if "${INGEST_SCRIPT}" \
        --repository "${REPO_NAME}" \
        --basedir "${VERSIONS_SUBPATH}" \
        "${local_tar}"; then
        echo "Ingest succeeded for ${filename}."
    else
        echo "ERROR: CVMFS ingestion failed for ${filename}." >&2
        # Keep the files for troubleshooting
        continue
    fi
```

Neves-P (changes requested):
Suggestions with updates accounting for `VERSIONS_SUBPATH` workflow in `ingest-tarball.sh`.

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1789 @@
...
+# a name for a dir in the bucket in which to archive tarballs, so that a subsequent run doesn't re-ingest them
+ARCHIVE_PREFIX="archive"
+# Ensure the download directory exists
+mkdir -p "${DOWNLOAD_DIR}"
+
+# We will leverage a script from eessi-bot-software-layer (for signature verification - optional)
+git clone https://github.com/EESSI/eessi-bot-software-layer.git
+
+# We will leverage a script from filesystem-layer (for tarball ingestion)
+git clone https://github.com/EESSI/filesystem-layer.git
```
bedroge, line 1218: Having these here will break things for a second run, as these dirs will already exist. We should either add a check, or maybe just let people do this manually and add a variable for the paths to these dirs at the top? I think I'd prefer the latter, makes it a bit more flexible in case you want to (temporarily) modify these scripts, e.g. for testing.

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1789 @@
...
+    echo "Downloading tarball signature file... s3://${BUCKET}/${sig_key} to ${local_tar_sig}"
+    aws s3 cp "s3://${BUCKET}/${sig_key}" "${local_tar_sig}"
+
+    if [ $? -eq 0 ]; then
+        echo "Tarball signature file downloaded."
+    else
+        echo "WARNING: Failed to download tarball signature file. Continuing to next tarball (not ingesting ${filename})."
+        # No point in continuing this loop iteration, we'll fail the signature verification check anyway
+        continue
+    fi
```
bedroge, line 1323: This basically does mean that it's not optional, as it will now entirely skip tarballs with a signature file. Maybe we should have another variable at the top that allows people to enable/disable the signature checking? If it's enabled but the sig file(s) cannot be downloaded -> hard error for that tarball. If it's disabled, don't even try.

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1789 @@
...
+
+In the example commands below, we assume that the current environment is set up:
+
+``` { .bash .copy }
+#!/bin/bash
+
+IFS=$'\n\t'         # sane field splitting
+BUCKET="<bucket_name>"
+DOWNLOAD_DIR=/prefix/for/tarball/staging  # Some directory to temporarily store tarballs on the Stratum 0
+ALLOWED_SIGNERS=/path/to/allowed/signers/file  # Optional, needed in step 4
```
Neves-P:
```suggestion

# We will leverage a script from filesystem-layer (for tarball ingestion)
# Do this beforehand while setting up, not as part of this script
# mkdir -p /path/to/eessi/gh/repos/filesystem-layer && cd /path/to/eessi/gh/repos/filesystem-layer
# git clone https://github.com/EESSI/filesystem-layer.git

# We will leverage a script from eessi-bot-software-layer (for signature verification - optional)
# Do this beforehand while setting up, not as part of this script
# mkdir -p /path/to/eessi/gh/repos/eessi-bot-software-layer && cd /path/to/eessi/gh/repos/eessi-bot-software-layer
# git clone https://github.com/EESSI/eessi-bot-software-layer.git

# Set the prefixes used above to be able to leverage those scripts
BOT_PREFIX=/path/to/eessi/gh/repos/eessi-bot-software-layer
FILESYSTEM_LAYER_REPO_PREFIX=/path/to/eessi/gh/repos/filesystem-layer

ALLOWED_SIGNERS=/path/to/allowed/signers/file  # Optional, needed in step 4
```

docs/site_build/site_cvmfs.md
Neves-P:
```suggestion
```

Moved this step as comments above. I think I can also add something in the text, or this will be easy to miss.

docs/site_build/site_cvmfs.md
Neves-P:
```suggestion
# ALLOWED_SIGNERS=/path/to/allowed/signers/file  # Optional, needed in step 4
```

Comment this out, we will check if the variable has been defined later to know if we will care about signing tarballs or not.

docs/site_build/site_cvmfs.md
```diff
@@ -0,0 +1,1789 @@
...
+
+    # ---- Download metadata signature file ----
+    echo "Downloading metadata signature file... s3://${BUCKET}/${meta_sig_key} to ${local_meta_sig}"
+    aws s3 cp "s3://${BUCKET}/${meta_sig_key}" "${local_meta_sig}"
+    if [ $? -eq 0 ]; then
+        echo "Metadata signature file downloaded."
+    else
+        echo "WARNING. Failed to download metadata signature file. Continuing to next tarball (not ingesting ${filename})."
+        continue
+    fi
```
Neves-P:
```suggestion
    if [ -z "$ALLOWED_SIGNERS" ]; then
        echo "$ALLOWED_SIGNERS is not set, skipping signature step."
    else
        # Full local paths
        local_tar_sig="${DOWNLOAD_DIR}/${sig_file}"
        local_meta_sig="${DOWNLOAD_DIR}/${meta_sig_file}"

        # Remote paths
        sig_key=${key}.sig
        meta_sig_key=${key}.meta.txt.sig
        
        # ---- Download tarball signature file (optional) ----
        echo "Downloading tarball signature file... s3://${BUCKET}/${sig_key} to ${local_tar_sig}"
        aws s3 cp "s3://${BUCKET}/${sig_key}" "${local_tar_sig}"

        if [ $? -eq 0 ]; then
            echo "Tarball signature file downloaded."
        else
            echo "WARNING: Failed to download tarball signature file. Continuing to next tarball (not ingesting ${filename})."
            # No point in continuing this loop iteration, we'll fail the signature verification check anyway
            continue
        fi
    
        # ---- Download metadata signature file ----
        echo "Downloading metadata signature file... s3://${BUCKET}/${meta_sig_key} to ${local_meta_sig}"
        aws s3 cp "s3://${BUCKET}/${meta_sig_key}" "${local_meta_sig}"
        if [ $? -eq 0 ]; then
            echo "Metadata signature file downloaded."
        else
            echo "WARNING. Failed to download metadata signature file. Continuing to next tarball (not ingesting ${filename})."
            continue
        fi
    fi
```

Neves-P: A blunt skip of this whole chunk if we don't have `$ALLOWED_SIGNERS` set in the first place

Neves-P (changes requested):
I haven't tested this in the wild yet, but here's a simple proposal for handling cases when we don't want to sign the files. It ignores that step if `$ALLOWED_SIGNERS` is not set and writes a message saying so. If the file is there, it carries on with the signature workflow as usual.

docs/site_build/site_cvmfs.md
Neves-P: Here we also point admins to cloning the filesystem-layer and bot repo beforehand.

docs/site_build/site_cvmfs.md
Neves-P: I tried addressing this in the last batch of suggestions.

PR #812, opened 2026-07-07
Update documentation for MUSICA

Neves-P:
Removed note about MUSICA being in an open test phase as the system is now in production. See https://asc.ac.at/news/2026/news-records/supercomputer-musica-goes-into-operation/


ocaisa (approved):
LGTM

PR #816, opened 2026-07-10
Update CUDA configuration guide

jngrad:
Description of changes:
- fix CUDA installation instructions
- document common issues and workarounds
- add a "copy to clipboard" widget to bash scripts

### DIFF
--- a/docs/site_specific_config/gpu.md
+++ b/docs/site_specific_config/gpu.md
@@ -43,14 +43,18 @@ to store the symlinks to the drivers. For example, to store these under `/opt/ee
 sudo bash -c "echo 'EESSI_NVIDIA_OVERRIDE_DEFAULT=/opt/eessi/nvidia' >> /etc/cvmfs/default.local"
 ```
 
+Keep in mind that values stored in `/etc/cvmfs/default.local` can be overridden by values stored in `/etc/cvmfs/domain.d/eessi.io.local`.
+
 *Step 2 (EESSI 2023.06, optional):* Change the location in which the symlinks will end up by configuring `EESSI_HOST_INJECTIONS` explicitly (default: `/opt/eessi`):
 
-```{ .bash copy }
+```{ .bash .copy }
 sudo bash -c "echo 'EESSI_HOST_INJECTIONS=/desired/path/to/host/injections' >> /etc/cvmfs/default.local"
 ```
 
+Keep in mind that values stored in `/etc/cvmfs/default.local` can be overridden by values stored in `/etc/cvmfs/domain.d/eessi.io.local`.
+
 *Step 3:* To actually reconfigure the variant symlinks, reload the updated CernVM-FS configuration using:
-```{ .bash copy }
+```{ .bash .copy }
 sudo cvmfs_config reload software.eessi.io
 ```
 
@@ -66,7 +70,7 @@ sudo cvmfs_config reload software.eessi.io
 
 !!! tip "Maintaining different driver versions for each EESSI version"
     The standard approach for EESSI >= 2025.06 means that the drivers may be found by any EESSI version. If you prefer to create one set of symlinks per EESSI
-    version, instead of defining a single location through EESSI_NVIDIA_OVERRIDE_DEFAULT, you can define one per EESSI version, by setting EESSI_<VERSION>_NVIDIA_OVERRIDE.
+    version, instead of defining a single location through `EESSI_NVIDIA_OVERRIDE_DEFAULT`, you can define one per EESSI version, by setting `EESSI_<VERSION>_NVIDIA_OVERRIDE`.
     For example:
     ```{ .bash .copy}
     sudo bash -c "echo 'EESSI_202506_NVIDIA_OVERRIDE=/opt/eessi/2025.06/nvidia' >> /etc/cvmfs/default.local"
@@ -96,6 +100,10 @@ Then, change to the correct directory:
 - For EESSI 2023.06: `/cvmfs/software.eessi.io/host_injections/${EESSI_VERSION}/compat/${EESSI_OS_TYPE}/${EESSI_CPU_FAMILY}/lib`
 
 Then, manually create the symlinks for each of the files in the aforementioned list (if they exist on your system) to the current directory.
+You can use the command `locate libcuda.so` to determine the absolute path of the libraries on your system.
+Sometimes, these libraries have versioned variants (e.g. with extension `.so.1`), which need to be linked as well.
+If a CUDA application fails at runtime due to a missing shared library, you can determine which one
+is missing with `compute-sanitizer --tool memcheck ./myapp` (requires `module load CUDA/<cuda_version>`).
 
 #### Runtime support when using EESSI in a container {: #nvidia_eessi_container } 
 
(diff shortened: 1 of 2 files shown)


docs/site_specific_config/gpu.md
```diff
@@ -66,7 +70,7 @@ sudo cvmfs_config reload software.eessi.io
 
 !!! tip "Maintaining different driver versions for each EESSI version"
     The standard approach for EESSI >= 2025.06 means that the drivers may be found by any EESSI version. If you prefer to create one set of symlinks per EESSI
-    version, instead of defining a single location through EESSI_NVIDIA_OVERRIDE_DEFAULT, you can define one per EESSI version, by setting EESSI_<VERSION>_NVIDIA_OVERRIDE.
+    version, instead of defining a single location through `EESSI_NVIDIA_OVERRIDE_DEFAULT`, you can define one per EESSI version, by setting `EESSI_<VERSION>_NVIDIA_OVERRIDE`.
```
jngrad, line 73: Without backticks, the `<VERSION>` part of the environment variable name was interpreted as an HTML tag and was not displayed as text.

docs/site_specific_config/gpu.md
```diff
@@ -43,14 +43,18 @@ to store the symlinks to the drivers. For example, to store these under `/opt/ee
 sudo bash -c "echo 'EESSI_NVIDIA_OVERRIDE_DEFAULT=/opt/eessi/nvidia' >> /etc/cvmfs/default.local"
 ```
 
+Keep in mind that values stored in `/etc/cvmfs/default.local` can be overridden by values stored in `/etc/cvmfs/domain.d/eessi.io.local`.
+
 *Step 2 (EESSI 2023.06, optional):* Change the location in which the symlinks will end up by configuring `EESSI_HOST_INJECTIONS` explicitly (default: `/opt/eessi`):
 
-```{ .bash copy }
+```{ .bash .copy }
```
jngrad, line 50: The dot is necessary to create the CSS class that enables the "copy to clipboard" widget.

docs/site_specific_config/gpu.md
```diff
@@ -166,11 +178,13 @@ Thus, you may want to periodically run this script to pick up on new CUDA and cu
 
 The quickest way to test if software installations included in EESSI can access and use your GPU is to run the
 `deviceQuery` executable that is part of the `CUDA-Samples` module:
-```
+```{ .bash .copy }
 module load CUDA-Samples
 deviceQuery
```
jngrad, line 184: Will work with the 2025.06 stack once https://github.com/EESSI/software-layer/pull/1541 is merged.

docs/site_specific_config/gpu.md
```diff
@@ -133,9 +141,13 @@ sudo bash -c "echo 'EESSI_HOST_INJECTIONS=/my/custom/prefix' >> /etc/cvmfs/defau
 Third, run the helper script to install the CUDA and cuDNN versions that are used _in that version of EESSI_.
 
 ```{ .bash .copy }
-/cvmfs/software.eessi.io/versions/${EESSI_VERSION}/scripts/gpu_support/nvidia/install_cuda_and_libraries.sh
+/cvmfs/software.eessi.io/versions/${EESSI_VERSION}/scripts/gpu_support/nvidia/install_cuda_and_libraries.sh --accept-cuda-eula --accept-cudnn-eula
```
ocaisa: This one I think we should keep but then add an admonition that explains it will ask you for additional options. The reasoning is that if the user actively add these options then they can only blame themselves if they violate those agreements.

docs/site_specific_config/gpu.md
```diff
@@ -133,9 +141,13 @@ sudo bash -c "echo 'EESSI_HOST_INJECTIONS=/my/custom/prefix' >> /etc/cvmfs/defau
...
 ```
 
+Be sure to work in a clean environment and outside your local EasyBuild configuration.
+If the installation is interrupted by the error message "Files missing CUDA PTX code",
+re-run the command with environment variable `EASYBUILD_CUDA_SANITY_CHECK_ACCEPT_MISSING_PTX=1`.
```
ocaisa:
```suggestion
re-run the command with environment variable `EASYBUILD_CUDA_SANITY_CHECK_ACCEPT_MISSING_PTX=1` set.
```

docs/site_specific_config/gpu.md
ocaisa: That looks like it will take a little longer that hoped...but I will get there!

ocaisa (changes requested):

docs/site_specific_config/gpu.md
```diff
@@ -43,14 +43,18 @@ to store the symlinks to the drivers. For example, to store these under `/opt/ee
 sudo bash -c "echo 'EESSI_NVIDIA_OVERRIDE_DEFAULT=/opt/eessi/nvidia' >> /etc/cvmfs/default.local"
 ```
 
+Keep in mind that values stored in `/etc/cvmfs/default.local` can be overridden by values stored in `/etc/cvmfs/domain.d/eessi.io.local`.
```
casparvl, line 46: I'm wondering if we should actually modify the instruction here alltogether, and say that they should do:

```
sudo bash -c "echo 'EESSI_NVIDIA_OVERRIDE_DEFAULT=/opt/eessi/nvidia' >> /etc/cvmfs/domain.d/eessi.io.local
```

I'm not sure why we ever decided to put this in `default.local`, but it doesn't make much sense. This variant symlink is only relevant for EESSI anyway, it should be in the EESSI-specific (local) config.

One could even argue if it should be under `config.d`, i.e. the repository-specific files. But... that might bite us later if we ever rename the repo or set up another repo under the same domain in which we want to reuse the same variant symlinks, etc. For now, I'd say `domain.d/eessi.io.local` is specific enough.

I'd like to get one other opinion on this though. @ocaisa what do you think?

docs/site_specific_config/gpu.md
```diff
@@ -43,14 +43,18 @@ to store the symlinks to the drivers. For example, to store these under `/opt/ee
...
+
 *Step 2 (EESSI 2023.06, optional):* Change the location in which the symlinks will end up by configuring `EESSI_HOST_INJECTIONS` explicitly (default: `/opt/eessi`):
 
-```{ .bash copy }
+```{ .bash .copy }
 sudo bash -c "echo 'EESSI_HOST_INJECTIONS=/desired/path/to/host/injections' >> /etc/cvmfs/default.local"
 ```
 
+Keep in mind that values stored in `/etc/cvmfs/default.local` can be overridden by values stored in `/etc/cvmfs/domain.d/eessi.io.local`.
```
casparvl, line 54: See previous comment.

docs/site_specific_config/gpu.md
```diff
@@ -96,6 +100,10 @@ Then, change to the correct directory:
 - For EESSI 2023.06: `/cvmfs/software.eessi.io/host_injections/${EESSI_VERSION}/compat/${EESSI_OS_TYPE}/${EESSI_CPU_FAMILY}/lib`
 
 Then, manually create the symlinks for each of the files in the aforementioned list (if they exist on your system) to the current directory.
+You can use the command `locate libcuda.so` to determine the absolute path of the libraries on your system.
```
casparvl, line 103: I'm not too familiar with `locate`, but this instruction might not very generically work. Not all systems have `locate` installed. Also, as far as I know, it searches a database, which may or may not have indexed the relevant path. So... your mileage may vary.

Not that I have a better alternative, but maybe we should phrase it a bit more careful (like 'if available, you can use a command like ...').

docs/site_specific_config/gpu.md
```diff
@@ -136,6 +144,11 @@ Third, run the helper script to install the CUDA and cuDNN versions that are use
 /cvmfs/software.eessi.io/versions/${EESSI_VERSION}/scripts/gpu_support/nvidia/install_cuda_and_libraries.sh
 ```
 
+The script may ask for additional command line options, e.g. to accept the terms of the EULA.
+Be sure to work in a clean environment and outside your local EasyBuild configuration.
+If the installation is interrupted by the error message "Files missing CUDA PTX code",
+re-run the command with environment variable `EASYBUILD_CUDA_SANITY_CHECK_ACCEPT_MISSING_PTX=1` set.
```
casparvl, line 150: Interesting. I don't have this exactly in my mind, but I guess the EasyBuild we load in `install_cuda_and_libraries.sh` uses the same config as `EESSI-extend` - and I'm guessing that set a `--cuda-compute-capabilities` value if you're on a GPU-enabled node?

I'm wondering if we should put this in the docs - if we tell people to do this any way (and I think it's reasonable...) maybe we better just adjust `install_cuda_and_libraries.sh` and make sure it gets set there (as environment variable, or even as a command line option).

docs/site_specific_config/gpu.md
ocaisa: If it is known to the OS, it should appear with `ldconfig -p | grep cuda`

jngrad:
@casparvl @ocaisa feel free to push directly to my branch if you see further opportunities for improvement :blush:

docs/site_specific_config/gpu.md
ocaisa: I've fixed this directly in the easystack file, this hack should not be necessary now.

ocaisa:
I was waiting to get the CUDA-Samples for 2025.06 merged before looking at this. That just happened today so I might try to finish this later

docs/site_specific_config/gpu.md
ocaisa: The needed PR is now merged

PR #835, opened 2026-08-07
disable `macros` and `inject_ld_json` plugins outside of CI

### LINKED ISSUE #834: very long build times due to mkdocs plugins
on my laptop, removing `macros` and `inject_ld_json` reduces the build time from >6 minutes to 35 seconds

smoors:
fixes https://github.com/EESSI/docs/issues/834

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -106,6 +106,7 @@ plugins:
   # see also https://mkdocs-macros-plugin.readthedocs.io/en/latest/macros/
   - macros:
       module_name: docs/available_software/macros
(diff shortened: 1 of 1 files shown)


ocaisa:
Timings in CI are consistent so I think this is working (assuming it results in a faster local build, I be no access to a laptop right now so can't check)

ocaisa (approved):
LGTM

ocaisa:
Just to note, there is a downside to this, the "Available software" page does not get created properly and you end up with a lot of weird looking code on that page.

PR #863, opened 2026-09-09
Add premier and fireflyk3 to the list of systems where EESSI is available

julianmorillo:


boegel:
@julianmorillo You're removing all references to `RISC-V` with these changes, we should probably keep that, and mention that these different nodes are partitions in a `RISC-V` Slurm cluster hosted at BSC?

julianmorillo:
@boegel  I removed the references because they referred to `riscv.eessi.io` which is discontinued. But I agree that mentioning that they are `RISC-V` clusters is convenient. I'll look into it.

boegel (approved):

PR #869, opened 2026-09-14
add ALICE HPC Cluster Leiden University to list of systems that provide EESSI

peterstol:
Add ALICE HPC Cluster Leiden University


boegel (approved):
lgtm

